PR #16, opened 2021-03-18
fixed docker issues

anbucher:
Docker is working now, I added also a docker-compose file. Please note that I included there the url to my own repository, this needs to be changed, perhaps could be outsourced to an .env-file.


orangecoding:
Hi,

thank you so much. There are some merge conflicts. Theoretically it would be nice if you could fix them but as we have spoken via email, I'll take care of this and squash-merge it into Fredy.

PR #42, opened 2022-01-25
Allow visiting the original provider URL

joschi:
Instead of truncating the original URL of each provider in the job configuration to 60 characters and losing a lot of context information, put a link to the original URL in the provider table which can be opened directly to verify what is being scraped by Fredy.

**Before**
Screenshot before the change

**After**
Screenshot after the change


orangecoding:
Looks awesome. Thanks!

PR #44, opened 2022-01-26
Telegram: Use job name instead of ID and link in title

joschi:
This PR improves the message formatting of the Telegram notification adapter.

The Telegram notification adapter is currently using the job ID (`jobKey`) in its message which is just a random, unique string. This makes it harder than necessary for humans to identify which job a listing originated from:
Telegram notification with job ID

The changes in this PR try to resolve the actual job name in the Telegram notification adapter and fall back to the job ID (`jobKey`) if the job name couldn't be resolved.

Additionally, the link to the listing is being combined with the title to make more use of the screen estate.

**Before**
Telegram notification before the change

**After**
Telegram notification after the change


orangecoding:
👍🏼

PR #45, opened 2022-01-26
Add support for Immo Südwest Presse (immo.swp.de)

joschi:
This PR adds a provider for the real estate listings of Südwest Presse: https://immo.swp.de/


joschi:
This _might_ even work with https://immobilien.augsburger-allgemeine.de/ since they seem to be using the same software.

orangecoding:
👍🏼

PR #46, opened 2022-01-30
Make `requestDriver` more resilient to errors

joschi:
If the async request performed in `requestDriver.makeDriver()` fails, it would call the `callback` function with empty parameters but then continue the execution which can lead to the following error and crash of Fredy:
```
Error while trying to scrape data. Received error: Request failed with status code 504
/fredy/lib/services/requestDriver.js:25
    if (typeof result.data === 'object' && url.toLowerCase().indexOf('scrapingant') !== -1) {
                      ^

TypeError: Cannot read properties of undefined (reading 'data')
    at driver (/fredy/lib/services/requestDriver.js:25:23)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

### DIFF
--- a/lib/services/requestDriver.js
+++ b/lib/services/requestDriver.js
@@ -7,29 +7,28 @@ function makeDriver(headers = {}) {
   let cookies = '';
 
   return async function driver(context, callback) {
-    const url = context.url;
-    let result;
     try {
-      result = await axios({
+      const url = context.url;
+      const result = await axios({
         url,
         headers: {
           ...headers,
           Cookie: cookies,
         },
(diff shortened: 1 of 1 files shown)


orangecoding:
@joschi a test is failing. I don't think this has something to do with your change, but can you have a look?

joschi:
@orangecoding Looks like an intermittent error with [wg-gesucht.de](https://www.wg-gesucht.de/) which has been fixed by now.

Successful test for wg-gesucht.de

PR #47, opened 2022-02-09
Fix address of immonet

CarlAmbroselli:
The value at Immonet is `0.16 km • Etagenwohnung • Berlin Tiergarten`, hence instead of selecting the index 1 (which would be "Etagenwohnung"), we should pick the last element of the splitted array.


orangecoding:
@CarlAmbroselli there are test failures. would you mind fixing them before merging?

orangecoding:
nevermind, i fixed it.

PR #48, opened 2022-02-09
Add sqlite adapter

CarlAmbroselli:
Adds an sqlite adapter that stores the values of all listings locally for post-processing.

### DIFF
--- a/lib/notification/adapter/sqlite.js
+++ b/lib/notification/adapter/sqlite.js
@@ -1,0 +1,33 @@
+const { markdown2Html } = require('../../services/markdown');
+const Database = require('better-sqlite3');
+
+/**
+ * simply prints out the found data to the console
+ * @param serviceName e.g immowelt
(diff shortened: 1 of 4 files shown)


lib/notification/adapter/sqlite.js
```diff
@@ -0,0 +1,33 @@
+const { markdown2Html } = require('../../services/markdown');
+const Database = require('better-sqlite3');
+
+/**
+ * simply prints out the found data to the console
```
orangecoding, line 5: please update the comment. This is a copy & paste from the console adapter ;)

orangecoding (changes requested):
Other than that 1 comment, this looks good to me.

lib/notification/adapter/sqlite.js
orangecoding: I'll do it myself. next time, please make sure to cleanup when copy and paste ;)  thanks

Issue #49, opened 2022-03-05
Add mattermost adapter


bergo:
Add mattermost adapter to send new listings to team chat

PR #50, opened 2022-03-05
Add mattermost adapter (#49)

bergo:
Created a mattermost adpater, based on the telegram adapter with a table of listing results

### DIFF
--- a/lib/notification/adapter/mattermost.js
+++ b/lib/notification/adapter/mattermost.js
@@ -1,0 +1,59 @@
+const { markdown2Html } = require('../../services/markdown');
+const { getJob } = require('../../services/storage/jobStorage');
+const axios = require('axios');
+
+/**
+ * sends new listings to mattermost
+ * @param serviceName e.g immowelt
+ * @param newListings an array with newly found listings
+ * @param notificationConfig config of this notification adapter
+ * @param jobKey name of the current job that is being executed
+ * @returns {Promise<Void> | void}
+ */
+exports.send = ({ serviceName, newListings, notificationConfig, jobKey }) => {
+  const { webhook, channel } = notificationConfig.find((adapter) => adapter.id === 'mattermost').fields;
+  const job = getJob(jobKey);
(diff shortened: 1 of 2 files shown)


lib/notification/adapter/mattermost.js
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * sends new listings to mattermost
+ * @param serviceName e.g immowelt
+ * @param newListings an array with newly found listings
+ * @param notificationConfig config of this notification adapter
+ * @param jobKey name of the current job that is being executed
+ * @returns {Promise<Void> | void}
+ */
+exports.send = ({ serviceName, newListings, notificationConfig, jobKey }) => {
```
orangecoding, line 13: As far as I understood, this adapter sends messages per chat? If so, it is useful to use the telegram adapter, which shortens the results? Mattermost should not have the restriction of 30 chars no?

lib/notification/adapter/mattermost.js
```diff
@@ -0,0 +1,59 @@
...
+  readme: markdown2Html('lib/notification/adapter/mattermost.md'),
+  description: 'Fredy will send new listings to your mattermost team chat.',
+  fields: {
+    webhook: {
+      type: 'text',
+      label: 'Webhook',
+      description: 'The incoming webhook url',
+    },
+    channel: {
+      type: 'channel',
```
orangecoding, line 54: the type is allways text (or number, but in your case it's text)

orangecoding (changes requested):

orangecoding:
thanks for this pr, please check the 2 comments.

lib/notification/adapter/mattermost.js
bergo: Yes, thats true. I used it for better mobile view, but works without in the same way.  Changed it.

lib/notification/adapter/mattermost.js
bergo: Ok, changed to text-type

PR #52, opened 2022-04-04
added run-script-os for win compatability

### LINKED ISSUE #51: Windows does not understand "export"
**Describe the bug**
When running `yarn run prod` on a windows machine the command `export BUILD_DEV='false'` in package.json fails with `'export' is not recognized as an internal or external command`.

**Expected behavior**
Replacing all instances of `export` with `set` if running on a windows machine fixes the issue. I don't know how to do that in a dynamic way though.

sven-simonsen:
closes #51 
Win uses "set" instead of unix's "export"
To make it dynamic I followed @orangecoding 's advice and used run-script-os https://www.npmjs.com/package/run-script-os

I have only tested on windows if `yarn run prod` and `yarn run dev` still work.
I have not run the tests and I have not checked the dependency management is correct.
Also I forked the repo since I lacked access rights to the main repo...

Tests still green (obviously)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,12 @@
   "description": "[F]ind [R]eal [E]states [d]amn eas[y].",
   "scripts": {
     "start": "node index.js",
-    "dev": "yarn && export BUILD_DEV='true' && export NODE_ENV='development' && webpack serve --progress --color --config ./webpack.dev.js",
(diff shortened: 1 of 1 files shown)


orangecoding:
This works. Thanks a lot!

Issue #57, opened 2022-05-31
Telegram Adapter 'Too Many Requests'


daniel17903:
**Describe the bug**
I am using fredy the first time and created a new Job using the Telegram Adapter. Initially a lot of listings were found. In the log I see many HTTP 429 Responses like this:

```
[AxiosError: Request failed with status code 429] {
  code: 'ERR_BAD_REQUEST',
  config: {
    transitional: {
      silentJSONParsing: true,
      forcedJSONParsing: true,
      clarifyTimeoutError: false
    },
...
```

I think the issue is that there currently is no limit of how many messages are sent. The telegram Bots FAQ says "avoid sending more than one message per second" and  "your bot will not be able to send more than 20 messages per minute to the same group" (see https://core.telegram.org/bots/faq#my-bot-is-hitting-limits-how-do-i-avoid-this).

So I think there should be some delay in between requests to the telegram API and maybe also some kind of retry mechanism if sending a message fails.

orangecoding:
Thanks, this commit should address the issue:
https://github.com/orangecoding/fredy/commit/bc018064218d82bd5c5a1e6338ddd8e4659cf5da

PR #58, opened 2022-06-03
Delay Telegram message sending to prevent HTTP 429 responses

daniel17903:
In order to prevent HTTP 429 responses the telegram adapter should not send more than 1 message per second and not more than 20 messsages per minute (see https://core.telegram.org/bots/faq#my-bot-is-hitting-limits-how-do-i-avoid-this).
#57

### DIFF
--- a/lib/notification/adapter/telegram.js
+++ b/lib/notification/adapter/telegram.js
@@ -15,6 +15,40 @@ const arrayChunks = (inputArray, perChunk) =>
     return all;
   }, []);
 
+const messageHeader = (jobName, serviceName, newListingsCount) => 
+  `<i>${jobName}</i> (${serviceName}) found <b>${newListingsCount}</b> new listings:\n\n`;
+
+const chunkToHtml = (chunk) => chunk.map(
+  (o) =>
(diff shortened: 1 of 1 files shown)


orangecoding:
@daniel17903 Thanks for this pr.

I however see the use of those kind of delays a little bit sceptical. Because of the single threaded nature of javascript, blocking this is potentially blocking Fredy completely. Depending on how short the crawling interval has been set, we could potentially end up in a situation, where we wait longer than it takes for the next run.

I would have a different proposal.

First of, I have checked again the max size of a message and notice, it has been changed to 4096 chars (it was half the size before), thus I would go up with the number of entries within a chunk. With the messages we are sending, we can easily go up to 8  entries per chunk (potentially more) and we would still be under 4kb.

Next I would listen to the HTTP response instead of slowing down the whole process directly. If we are hitting 429, we wait a second and retry. The retry happens a max of 3 times with exponential backoff (maybe up to 10 seconds) and if it still doesn't work, we log this as an error.

I am pretty sure with these changes, we will never end up in a situation, where we send too many messages a minute. (Or better; the ONLY situation where something like this might happen would be if we crawl some page the first time, thus have a bunch of results)

What do you think?

PR #59, opened 2022-06-10
Scrapingant proxies

orangecoding:
This pr introduces a new ui setting to chose between ScrapingAnt Datacenter & Residential proxies. The retry mechanism will now take this new setting into account.

Also the new requestDriver will not fail on 404 / 429 status.


kami4ka:
Looks good to me! Thanks!

PR #65, opened 2022-12-17
Provider for immobilien.de

vanHekthor:
This PR adds a provider for [immobilien.de](https://www.immobilien.de/).


orangecoding:
nice, I didn't even know about them. Thanks 💋

PR #66, opened 2022-12-17
Suggestions to improve the README information

vanHekthor:
2 changes that I think would make it easier for users and developers to understand Fredy:

### 1. Fix the mix-up of providers and adapters
The descriptions of providers and adapters seem to have been mixed up.
When comparing them to the project structure (the adapter and provider folders), it clearly is supposed to be the other way around.
It is also stated the other way around in the `CONTRIBUTING.md`.

### 2. Provide commands for development
The following section is a bit confusing in my opinion:
> To run _Fredy_ in development mode, you need to run the backend & frontend separately. Run the backend in your favorite IDE, the frontend can be started from the terminal.

The way it is phrased implies that you need an IDE to run the backend whereas the frontend can be started only using the terminal.
Just providing the 2 commands (`yarn run start` and `yarn run dev`) for running the backend and the frontend may be more straightforward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,20 +33,20 @@ _Fredy_ will start with the default port, set to `9998`. You can access _Fredy_ 
 ## Understanding the fundamentals
 There are 3 important parts in Fredy, that you need to understand to leverage the full power of _Fredy_.
 
-#### Adapter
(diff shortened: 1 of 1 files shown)


orangecoding:
Cool, thanks :)

PR #75, opened 2023-04-06
Fix validation and add ntfy as notification adapter

dalins:
Add ntfy as notification adapter and there was a bug inside the validation of number input which is fixed. This fix was necessary to enable ntfy priority.

### DIFF
--- a/lib/notification/adapter/ntfy.js
+++ b/lib/notification/adapter/ntfy.js
@@ -1,0 +1,50 @@
+import { markdown2Html } from '../../services/markdown.js';
+import { getJob } from '../../services/storage/jobStorage.js';
+import fetch from 'node-fetch';
+import utf8 from 'utf8';
+export const send = ({ serviceName, newListings, notificationConfig, jobKey }) => {
+  const { priority, server, topic } = notificationConfig.find((adapter) => adapter.id === 'ntfy').fields;
+  const job = getJob(jobKey);
+  const jobName = job == null ? jobKey : job.name;
+  const url = `${server}/${topic}`;
+  const promises = newListings.map((newListing) => {
+    const message = utf8.encode(
+      `Address: ${newListing.address} Size: ${newListing.size.replace(/2m/g, '$m^2$')} Price: ${newListing.price}`
+    );
+    return fetch(url, {
+      method: 'POST',
+      headers: {
+        'X-Click': newListing.link,
+        'X-Message': message,
+        'X-Priority': priority,
+        'X-Tags': [serviceName, jobName],
+        'X-Title': utf8.encode(newListing.title),
+      },
+    });
+  });
+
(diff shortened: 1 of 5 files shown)


ui/src/views/jobs/mutation/components/notificationAdapter/NotificationAdapterMutator.jsx
```diff
@@ -25,9 +25,12 @@ const validate = (selectedAdapter) => {
       results.push('All fields are mandatory and must be set.');
       continue;
     }
-    if (uiElement.type === 'number' && (typeof uiElement.value !== 'number' || uiElement.value < 0)) {
-      results.push('A number field cannot contain anything else and must be > 0.');
-      continue;
+    if (uiElement.type === 'number') {
+      const numberValue = parseFloat(uiElement.value);
```
orangecoding, line 29: why `parseFloat` wouldn't `parseInt` be more sufficient?

lib/notification/adapter/ntfy.js
```diff
@@ -0,0 +1,50 @@
...
+    },
+    server: {
+      type: 'text',
+      label: 'Server-URL',
+      description: 'The server url to the send the notification to.',
+    },
+    topic: {
+      type: 'text',
+      label: 'topic',
+      description: 'The topic where fredy should send notifications to.',
```
orangecoding, line 47: I'd suggest to add the following:
```
The topic where fredy should send notifications to. The topic is a secret, only known to you, make sure it is something not generic.
```

lib/notification/adapter/ntfy.js
```diff
@@ -0,0 +1,50 @@
...
+export const send = ({ serviceName, newListings, notificationConfig, jobKey }) => {
+  const { priority, server, topic } = notificationConfig.find((adapter) => adapter.id === 'ntfy').fields;
+  const job = getJob(jobKey);
+  const jobName = job == null ? jobKey : job.name;
+  const url = `${server}/${topic}`;
+  const promises = newListings.map((newListing) => {
+    const message = utf8.encode(
+      `Address: ${newListing.address} Size: ${newListing.size.replace(/2m/g, '$m^2$')} Price: ${newListing.price}`
+    );
+    return fetch(url, {
```
orangecoding, line 14: while I totally appreciate not to use a lib for a simple http call, in this case, it might make sense to use one.

I'm thinking about using this: https://github.com/cityssm/node-ntfy-publish

If you decide against this, at least please don't use utf8 as an external dependency, coz it's unmaintained (last update 6 years ago and tons of issues)

orangecoding (changes requested):
Thanks so much for your contribution. I did not even know about ntfy... 😱 
Could you check the comments I made?

orangecoding (approved):

PR #76, opened 2023-04-15
fix Mailjet adapter

janekbettinger:


janekbettinger:
The MailJet API has changed, see e.g. https://github.com/mailjet/mailjet-apiv3-nodejs#api-setup and https://stackoverflow.com/a/74869217.

In addition, the case of `adapter.id` did not match the one used in the config.

Successfully tested in my local setup.

orangecoding:
Thank you

PR #82, opened 2023-10-05
fix: kleinanzeigen price

jstnw10:
Hey @orangecoding,

first of all, thank you so much for maintaining this project :)

Here is a quick fix to get the price of a Kleinanzeigen listing again.

Cheers


orangecoding:
Awesome. Thanks a ton :)

PR #100, opened 2024-07-11
Update scrapping ant error handling

ivanovvladi:
Thanks for creating and maintaining this helpful tool

I noticed that `scrapping ant` sends 423 code with a message 

> Our browser was detected by target site. Retry the request or try adjusting proxy country, proxy type and browser rendering settings. Our documentation: https://docs.scrapingant.com/ and support email: support@scrapingant.com

and a request doesn't fall into the catch -> retry block. Hence, I'm opening this PR to update the error handling

### DIFF
--- a/lib/services/requestDriver.js
+++ b/lib/services/requestDriver.js
@@ -24,13 +24,16 @@ function makeDriver(headers = {}) {
         },
       });
       const result = await response.text();
+      if (EXPECTED_STATUS_CODES.includes(response.status)) {
+        throw new Error(`${response.status}`);
+      }
       if (cookies.length === 0) {
         cookies = response.headers.raw()['set-cookie'] || [];
       }
       callback(null, result);
     } catch (exception) {
       /* eslint-disable no-console */
-      if (!EXPECTED_STATUS_CODES.includes(exception.response?.status)) {
+      if (!EXPECTED_STATUS_CODES.includes(exception.response?.status) && !EXPECTED_STATUS_CODES.includes(Number(exception.message))) {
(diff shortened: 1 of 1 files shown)


lib/services/requestDriver.js
```diff
@@ -24,13 +24,16 @@ function makeDriver(headers = {}) {
...
+        throw new Error(`${response.status}`);
+      }
       if (cookies.length === 0) {
         cookies = response.headers.raw()['set-cookie'] || [];
       }
       callback(null, result);
     } catch (exception) {
       /* eslint-disable no-console */
-      if (!EXPECTED_STATUS_CODES.includes(exception.response?.status)) {
+      if (!EXPECTED_STATUS_CODES.includes(exception.response?.status) && !EXPECTED_STATUS_CODES.includes(Number(exception.message))) {
```
orangecoding, line 36: If the status message is not a number, this would generate an exception no?

orangecoding (changes requested):

lib/services/requestDriver.js
ivanovvladi: Thanks for reviewing the PR. That was my first impression, too; however, it returns
```js
    /**
     * A value that is not a number.
     * In equality comparisons, NaN does not equal any value, including itself. To test whether a value is equivalent to NaN, use the isNaN function.
     */
    readonly NaN: number;
```

https://github.com/microsoft/TypeScript/blob/6d3be985c82bead3b41348de76efec8110c677c5/src/lib/es5.d.ts#L585

and the `if-else` block is evaluated as expected. Since I'm not fully proficient with JS, there might be development standards that defer from the implementation. Let me know if that's the case, and I'd be happy to change it.

PR #106, opened 2024-08-11
Update the Telegram adapter to add a rich message option with a photo

ivanovvladi:
That PR updates the Telegram adapter and adds a rich message option with a photo.

It also updates a predefined example.

<table>
  <tr>
    <th>Plain message</th> 
    <th>Rich message</th>
  </tr>
  <tr>
    <td></td> 
    <td></td> 
  </tr>
</table>

**Note:** the PR adds image parsing for immobilienscout24, kleinanzeigen, 1a-immobilienmarkt.

### DIFF
--- a/lib/notification/adapter/telegram.js
+++ b/lib/notification/adapter/telegram.js
@@ -15,10 +15,83 @@ const arrayChunks = (inputArray, perChunk) =>
     all[ch] = [].concat(all[ch] || [], one);
     return all;
   }, []);
-function shorten(str, len = 30) {
+
+function shorten(str, len = 45) {
   return str.length > len ? str.substring(0, len) + '...' : str;
 }
-export const send = ({ serviceName, newListings, notificationConfig, jobKey }) => {
+
+function isValidURL(url) {
+  try {
+    new URL(url);
+    return true;
+  } catch (_) {
+    return false;
+  }
+}
+
+const sendTelegramMessage = (token, chatId, message, isPhoto = false, photoUrl = '') => {
+  const url = isPhoto
+    ? `https://api.telegram.org/bot${token}/sendPhoto`
+    : `https://api.telegram.org/bot${token}/sendMessage`;
+  const body = isPhoto
+    ? JSON.stringify({
+        chat_id: chatId,
+        photo: photoUrl,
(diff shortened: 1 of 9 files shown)


lib/notification/adapter/telegram.js
```diff
@@ -15,49 +15,111 @@ const arrayChunks = (inputArray, perChunk) =>
...
+  try {
+    new URL(url);
+    return true;
+  } catch (_) {
+    return false;
+  }
+}
+
+const sendTelegramMessage = (token, chatId, message, isPhoto = false, photoUrl = '') => {
+  const url = isPhoto
```
orangecoding, line 33: 1. I think a simple if/else would be easier to read here..
2. The api endpoint /sendPhoto works like /sendMessage, but it can inlcude a photo url?

lib/provider/kleinanzeigen.js
```diff
@@ -25,6 +25,7 @@ const config = {
     link: '.aditem-main .text-module-begin a@href | removeNewline | trim',
     description: '.aditem-main p:not(.text-module-end) | removeNewline | trim',
     address: '.aditem-main--top--left | trim | removeNewline',
+    image: '.aditem-image .imagebox img@src',
```
orangecoding, line 28: Would you mind adding this for the other providers as well, at least where it is available?

orangecoding (changes requested):

orangecoding:
@ivanovvladi Thanks mate, I have added 2 comments. Sorry it took a while..

lib/notification/adapter/telegram.js
ivanovvladi: Thanks for the suggestion, @orangecoding; I replaced the ternary operator. As for the API calls, there are some differences, but for the scope of this project, you're absolutely right - they work identically with only photo URL disparity

orangecoding:
@ivanovvladi have you read my second comment about adding the photo to all provider?

ivanovvladi:
> @ivanovvladi have you read my second comment about adding the photo to all provider?

Hi, @orangecoding; thanks for the heads-up; I missed the comment. I will update leftover providers as soon as I get time

Issue #102, opened 2024-08-07
Feature Request: Detect changed prices


dekiesel:
**Is your feature request related to a problem? Please describe.**
-

**Describe the solution you'd like**
There was a house I was interested in, but it had a damaged basement. Fredy found the listing since it was in my price range. I decided to pass on the house because `price+basement repair>budget`. 
Today I saw that they dropped the price by 100k and suddenly the house was within my budget, but I didn't see the new price until after it was sold. 

tl;dr: I'd like to get a notification when a listing has a price change.

**Describe alternatives you've considered**
Manually checking every listing, but that kinda defeats the purpose of fredy.

orangecoding:
Uhh.. cool feature. The problem of this is I don't store the actual data of a listing I sent, but rather only the listing id to make sure I don't send listings twice. For this to work, I would need to store specific props like price...

Are there ANY other props you could think of that might justify sending a listing again?

dekiesel:
Hmmm, that's a great question. Maybe square meters (in case there was a typo or something)? But I doubt that happens often if ever. Or if the realtor cut goes up or down?

It might be an idea to just save everything (that's not deemed a duplicate listing) if you touch the code anyway so that it is easier to get alerts for any changes in case somebody comes up with a good use case

orangecoding:
I think I have a better idea. Rather than storing the listing id, I could generate a hash that includes the listing id and props like the price. If any of which changes, the hash changes, thus the listing is being sent again.

Wdyt?

dekiesel:
May I ask why you'd rather not store the ID? I'd personally like it if all the details were saved because I wanted to do some analytics with the data anyway until I realized that not everything gets saved.

The hash would definitely work though. Only issue is see is that it only indicates that *something* has changed. In case somebody else comes around with a different use case the hash won't be enough.

orangecoding:
You are right, it only indicates something has changed, not what has changed. The reason I don't want to store everything is - it is a lot of data (most of people I know of running Fredy on a Pi)
- it changes the structure of how I store data internally and I do not have the capacity at the moment to do those big changes...

What kind of analytics do you want to do?

dekiesel:
I don't think it's a lot of data, not even for a pi. You might not be able to do analytics on it (depending on the query you want to run), but I'd be surprised if a pi couldn't easily save this amount of data per listing.

[...]

PR #111, opened 2024-09-27
einsAImmobilien: get inner text only

dekiesel:
As discussed in [102](https://github.com/orangecoding/fredy/issues/102)

This *should* only get the text of the diff. I tested it by changing my fredy instance and waited for the next job. No errors. I tried running the tests, but it only printed `  #einsAImmobilien testsuite()`and then did nothing for 10m.

As I said, I know next to nothing about js, so just because there isn't an error doesn't mean it positively works. 

If you feel the quality of this PR is poor just feel free to delete it.

### DIFF
--- a/lib/provider/einsAImmobilien.js
+++ b/lib/provider/einsAImmobilien.js
@@ -23,7 +23,7 @@ const config = {
   sortByDateParam: 'sort_type=newest',
   crawlFields: {
     id: '.inner_object_data input[name="marker_objekt_id"]@value | int',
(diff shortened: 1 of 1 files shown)


orangecoding:
Hey..

Thanks, but rather than your change (which would break the tests), I've pushed this regex as a solution. It is also fixing the actual problem ;)

https://github.com/orangecoding/fredy/commit/d3cb3a58810e41bbe11af47f45389c0259b4bba9

dekiesel:
Appreciate it :D Thanks!

PR #125, opened 2025-04-22
Add immoscout mobile API provider to avoid failing bot checks

libklein:
Currently, the Immoscout provider fails due to various bot protection mechanisms implemented on the [Immoscout web page](https://immoscout24.de). It seems to be a cat and mouse game at this point that isn't worth playing for a project like fredy. 

However, no such protection mechanisms seems to be implemented for the Immoscout mobile API, used by, e.g. the android app. Here, specifying the expected user agent seems to suffice. 

I've implemented an alternative Immoscout provider using the mobile API. It works by translating the same search URL format used by the existing Immoscout provider into the format expected by the mobile API and uses js-native fetch to make raw requests to the API.

I've tried to stick to the provider architecture as much as possible, but had to implement a custom `fetchListings` function to avoid using playwright. Happy to do further work on this if we're looking for a less hacky way of implementing the mobile API provider.

### DIFF
--- a/conf/config.json
+++ b/conf/config.json
@@ -1,1 +1,1 @@
-{"interval":"60","port":9998,"workingHours":{"from":"","to":""},"demoMode":false,"analyticsEnabled":null}
+{"interval":1,"port":9998,"workingHours":{"from":"","to":""},"demoMode":false,"analyticsEnabled":false}
--- a/conf/config.json
+++ b/conf/config.json
@@ -1,1 +1,1 @@
-{"interval":1,"port":9998,"workingHours":{"from":"","to":""},"demoMode":false,"analyticsEnabled":false}
+{"interval":"60","port":9998,"workingHours":{"from":"","to":""},"demoMode":false,"analyticsEnabled":null}
--- a/lib/FredyRuntime.js
+++ b/lib/FredyRuntime.js
@@ -26,7 +26,7 @@ class FredyRuntime {
       //modify the url to make sure search order is correctly set
       Promise.resolve(urlModifier(this._providerConfig.url, this._providerConfig.sortByDateParam))
         //scraping the site and try finding new listings
-        .then(this._getListings.bind(this))
+        .then(this._providerConfig.getListings?.bind(this) ?? this._getListings.bind(this))
         //bring them in a proper form (dictated by the provider)
         .then(this._normalize.bind(this))
         //filter listings with stuff tagged by the blacklist of the provider
--- a/lib/provider/immoscout-mobile.js
+++ b/lib/provider/immoscout-mobile.js
@@ -1,3 +1,45 @@
+/**
+ * ImmoScout provider using the mobile API to retrieve listings.
+ *
+ * The mobile API provides the following endpoints:
+ * - GET /search/total?{search parameters}: Returns the total number of listings for the given query
+ *   Example: `curl -H "User-Agent: ImmoScout24_1410_30_._" https://api.mobile.immobilienscout24.de/search/total?searchType=region&realestatetype=apartmentrent&pricetype=calculatedtotalrent&geocodes=%2Fde%2Fberlin%2Fberlin `
+ *
+ * - POST /search/list?{search parameters}: Actually retrieves the listings. Body is json encoded and contains
+ *   data specifying additional results (advertisements) to return. The format is as follows:
+ *   ```
+ *   {
+ *   "supportedResultListTypes": [],
+ *   "userData": {}
+ *   }
+ *   ```
+ *   It is not necessary to provide data for the specified keys.
+ *
+ *   Example: `curl -X POST 'https://api.mobile.immobilienscout24.de/search/list?pricetype=calculatedtotalrent&realestatetype=apartmentrent&searchType=region&geocodes=%2Fde%2Fberlin%2Fberlin&pagenumber=1' -H "Connection: keep-alive" -H "User-Agent: ImmoScout24_1410_30_._" -H "Accept: application/json" -H "Content-Type: application/json" -d '{"supportedResultListType": [], "userData": {}}'`
+
+ * - GET /expose/{id} - Returns the details of a listing. The response contains additional details not included in the
+ *   listing response.
+ *
+ *   Example: `curl -H "User-Agent: ImmoScout24_1410_30_._" "https://api.mobile.immobilienscout24.de/expose/158382494"`
+ *
+ *
+ * It is necessary to set the correct User Agent (see `getListings`) in the request header.
+ *
+ * Note that the mobile API is not publicly documented. I've reverse-engineered
+ * it by intercepting traffic from an android emulator running the immoscout app.
+ * Moreover, the search parameters differ slightly from the web API. I've mapped them
+ * to the web API parameters by comparing a search request with all parameters set between
+ * the web and mobile API. The mobile API actually seems to be a superset of the web API,
+ * but I have decided not to include new parameters as I wanted to keep the existing UX (i.e.,
+ * users only have to provide a link to an existing search).
+ *
+ * Limitations:
+ * - The current implementation of this provider *does not* support non-rental properties,
+ *   although the same approach can be used to implement support. It's just a matter of
+ *   mapping the web search URL to the corresponding mobile API URL.
+ * - Pagination support is not implemented.
+ */
+
 import utils, {buildHash} from '../utils.js';
 import queryString from 'query-string';
 let appliedBlackList = [];
@@ -68,7 +110,7 @@ export const init = (sourceConfig, blacklist) => {
   appliedBlackList = blacklist || [];
 };
 export const metaInformation = {
-  name: 'Immoscout (Mobile API)',
+  name: 'Immoscout',
   baseUrl: 'https://www.immobilienscout24.de/',
   id: 'immoscout-mobile',
 };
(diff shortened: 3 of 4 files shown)


conf/config.json
```diff
@@ -1 +1 @@
-{"interval":"60","port":9998,"workingHours":{"from":"","to":""},"demoMode":false,"analyticsEnabled":null}
\ No newline at end of file
+{"interval":1,"port":9998,"workingHours":{"from":"","to":""},"demoMode":false,"analyticsEnabled":false}
```
orangecoding, line 1: Can you revert that change?

lib/FredyRuntime.js
```diff
@@ -26,7 +26,7 @@ class FredyRuntime {
       //modify the url to make sure search order is correctly set
       Promise.resolve(urlModifier(this._providerConfig.url, this._providerConfig.sortByDateParam))
         //scraping the site and try finding new listings
-        .then(this._getListings.bind(this))
+        .then(this._providerConfig.getListings?.bind(this) ?? this._getListings.bind(this))
```
orangecoding, line 29: it is ok for now,  but needs refinement. Something I wanted to tackle anyway.

lib/provider/immoscout-mobile.js
```diff
@@ -0,0 +1,161 @@
+import utils, {buildHash} from '../utils.js';
```
orangecoding, line 1: Can you please add some documentation here?

Also the old immoscout implementation does not work anymore, thus, if this one works, it can replace the current implementation.

lib/provider/immoscout-mobile.js
```diff
@@ -0,0 +1,161 @@
...
+      userData: {}
+    })
+  });
+  if (!response.ok) {
+    console.error('Error fetching data from ImmoScout Mobile API:', response.statusText);
+    return [];
+  }
+
+  const responseBody = await response.json();
+  return responseBody.resultListItems.filter((item) => item.type === 'EXPOSE_RESULT').map(expose => {
```
orangecoding, line 23: interesting approach. I would have never believed this works.

lib/provider/immoscout-mobile.js
```diff
@@ -0,0 +1,161 @@
...
+      link: `${metaInformation.baseUrl}/expose/${item.id}`,
+      address: item.address?.line,
+    };
+  });
+}
+
+function nullOrEmpty(val) {
+  return val == null || val.length === 0;
+}
+function normalize(o) {
```
orangecoding, line 40: please run the linter, it will most likely find some flaws in here :)

lib/provider/immoscout-mobile.js
```diff
@@ -0,0 +1,161 @@
...
+  filter: applyBlacklist,
+  getListings: getListings
+};
+export const init = (sourceConfig, blacklist) => {
+  config.enabled = sourceConfig.enabled;
+  config.url = convertWebToMobile(sourceConfig.url);
+  appliedBlackList = blacklist || [];
+};
+export const metaInformation = {
+  name: 'Immoscout (Mobile API)',
```
orangecoding, line 71: I think you can just name it "Immoscout".

test/provider/immoscout-mobile.test.js
```diff
@@ -0,0 +1,34 @@
+import {expect} from 'chai';
+import {convertWebToMobile} from '../../lib/provider/immoscout-mobile.js';
+
+describe('#immoscout-mobile testsuite()', () => {
```
orangecoding, line 4: A good test case, however can you please make sure to properly test the provider (see the other provider tests, you can pretty much copy and paste the existing immoscout test)

orangecoding (changes requested):
Hey @libklein 
This is just awesome. I thought about various ways how to do that, but never thought about using the mobile api. Well done. I added some comments

test/provider/immoscout-mobile.test.js
libklein: Added a provider test based on the other providers. I've decided to split the test suite into the provider test and URL conversion unit tests to avoid issues with `provider.init(...)` and similarity cache (stateful) dependencies.

lib/FredyRuntime.js
libklein: Sounds good. Out of curiosity, what did you have in mind? Maybe moving to a prototype design? Let providers extend a common prototype and overwrite the given methods if necessary?

lib/provider/immoscout-mobile.js
libklein: I've actually run `npm lint`, but nothing is that file is flagged.

libklein:
I've added a bit of documentation on the mobile API and, more importantly, how I've reverse engineered it.

There are a few limitations that I have not addressed yet:
* The Mobile API uses pagination. I only fetch results on the first page right now. It's pretty straightforward to include other pages, but I am unsure how to avoid crawling the entire history. Does fredy provide a mechanism to retrieve the most recently scraped item?
* I have not implemented support for non-rental properties.

I'd suggest reviewing/testing the PR. I can then replace the existing `immoscout.js` with the mobile one in a final commit.

lib/FredyRuntime.js
orangecoding: I'm not yet fully sure what the solution will be, but I was thinking about something similar than what you did..

lib/provider/immoscout-mobile.js
orangecoding: interesting.
`yarn lint` finds some errors (funny enough none of which caused by you)
als I'd suggest to run `yarn run format`.

Actually there should be a pre-commit hook...

orangecoding:
I've been on vacation, thus the late response. 

What do you mean by this?
> I have not implemented support for non-rental properties.

Is there a different api for non rental estates?

> The Mobile API uses pagination. 
That is fine for now, as long as it is sorted and the newest is first. Did you make sure it is sorted actually?

libklein:
> What do you mean by this?
> ...
> As far as I understood it, this means, currently it only works for renting an appartment OR house, correct? What would it take to do the same for buying an apartment/house? I'm struggling a bit as I'm on mac and when I try to intercept the traffic, I do get all kinds of bot protection shit.

It currently only works for apartments. Houses or any other property type the app supports (garage, temporary living, shared flats, ...) are *not* supported by the provider. Again, adding compatibility should be straightforward. It's really an exercise of querying using the app and web interface, comparing the parameters. We'll need to support different filters for each property type (garages for instance do not support many of the apartment filters, and I expect the return to change as well). 

I think we need to align on a scope here - I really don't believe that the additional complexity of supporting garages is worth it (unless someone raises an issue).

I'm also unsure how well it'll work with drawn search areas - have not tested that. 

> Is there a different api for non rental estates?

I assume, but have not mapped it. The app uses a different selector when looking for properties to buy.

> That is fine for now, as long as it is sorted and the newest is first. Did you make sure it is sorted actually?

Yes, it is sorted.

> The link you generate has a double /. See
> 'https://www.immobilienscout24.de//expose/158932331'

Fixed

> I think it would make sense to extract the utilities code from the actualy provider. E.g. all the mapping stuff. What do you think?

I guess you are referring to the API parameter mapping function. Would we be using it for any other provider? I guess it would allow unit-testing the provider-agnostic mapping function in addition to the provider-specific one.

orangecoding:
> > What do you mean by this?
> 
> > ...
> 
> > As far as I understood it, this means, currently it only works for renting an appartment OR house, correct? What would it take to do the same for buying an apartment/house? I'm struggling a bit as I'm on mac and when I try to intercept the traffic, I do get all kinds of bot protection shit.
> 
> 
> 
> It currently only works for apartments. Houses or any other property type the app supports (garage, temporary living, shared flats, ...) are *not* supported by the provider. Again, adding compatibility should be straightforward. It's really an exercise of querying using the app and web interface, comparing the parameters. We'll need to support different filters for each property type (garages for instance do not support many of the apartment filters, and I expect the return to change as well). 
> 
> 
> 
> I think we need to align on a scope here - I really don't believe that the additional complexity of supporting garages is worth it (unless someone raises an issue).
> 
> 
> 
> I'm also unsure how well it'll work with drawn search areas - have not tested that. 
> 
> 
> 
> > Is there a different api for non rental estates?
> 
> 
> 
> I assume, but have not mapped it. The app uses a different selector when looking for properties to buy.
> 
> 
> 
> > That is fine for now, as long as it is sorted and the newest is first. Did you make sure it is sorted actually?
> 
> 
> 
> Yes, it is sorted.
> 
> 
> 
> > The link you generate has a double /. See
> 
> > 'https://www.immobilienscout24.de//expose/158932331'
> 
> 
> 
> Fixed
> 
> 
> 
> > I think it would make sense to extract the utilities code from the actualy provider. E.g. all the mapping stuff. What do you think?
> 
> 
> 
> I guess you are referring to the API parameter mapping function. Would we be using it for any other provider? I guess it would allow unit-testing the provider-agnostic mapping function in addition to the provider-specific one.

I do agree with the scoping. We can leave it for rental for now, but I think at least supporting buying apartments and houses should be implemented next to have a somewhat complete setup.
Would you be able to do it after we merge your pr?

Before merging your pr, I would suggest you remove the old immoscout provider and make the new one the standard as the old one doesn't work anyhow.

orangecoding:
@libklein I'll merge this now and do a couple of smaller changes. Would you be able to implement something to support purchases as well?

libklein:
I will look into it as soon as I get time :).

PR #128, opened 2025-05-19
ImmoScout: Web paths with SEO optimization to mobile params

leonbeon:
This PR adds the ability for the SEO optimized paths that the ImmoScout24 web app outputs to be converted to the usual query params that the mobile api uses.

These paths only appear when you use certain filters, and some of them supersede others. For example, most of the apartment type filters supersedes the balcony/garden filter, the latter is then represented as the usual query parameters. 

It also adds `newbuilding` to the allowed mobile params (for "Neubauwohnungen" filter"), and `handicappedaccessible` as an allowed equipment type (for "barrierefreie Wohnung" filter).

I tweaked the way equipment params are added to the mobileQuery variable to allow both sources of equipment types (web path conversion and the usual way via the actual params) to be merged.

Note that this only works for apartment renting for now. I assume the path structure is similar for houses and buying, but I haven't tested it. Maybe it would be better in the long run to write a function to build the web path segments, because they seem to follow a similar schema.

### DIFF
--- a/lib/services/immoscout/immoscout-web-translater.js
+++ b/lib/services/immoscout/immoscout-web-translater.js
@@ -79,6 +79,7 @@ const PARAM_NAME_MAP = {
   geocoordinates: 'geocoordinates',
   shape: 'shape',
   sorting: 'sorting',
+  newbuilding: 'newbuilding'
 };
 
 const EQUIPMENT_MAP = {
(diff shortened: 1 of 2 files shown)


orangecoding (approved):

orangecoding:
@leonbeon Thanks, I'm merging this and create a new release out of it, but for the sake of completeness, would you mind also including house purchase in a new pr?

PR #140, opened 2025-07-05
Typescript

vncnt-dev:
Hi there,

First of all, thank you for this great project — I really enjoy working with it!

This PR migrates the codebase from JavaScript to TypeScript. I undertook this migration primarily for myself, as I wanted to better understand the codebase and ensure more type safety while working with it. However, I believe this change could also benefit the main project in several ways:

- Improved maintainability: Type annotations make the code easier to understand and reason about.
- Fewer bugs: TypeScript helps catch common errors at compile time, reducing runtime issues.
- Better developer experience: Features like autocompletion and inline documentation are greatly improved when using TypeScript.

If you feel that sticking with JavaScript is the better fit for the project right now, I completely respect that — I just thought this migration might be a helpful option down the line.

At the moment, I’ve been using Bun as the package manager, mainly because of its built-in tooling for TypeScript support. That said, if you’d prefer the project to remain compatible with Yarn, I’d be more than happy to put in the work to make that adjustment — just let me know!

**I also noticed that there is another large open PR pending**. I totally understand if that one has priority. I wasn't sure how active it currently is, but just in case this PR is of interest, I wanted to go ahead and submit it. I'm more than happy to wait if you decide to prioritize the other PR or if you'd like to revisit this one later.

Let me know if there's anything I should adjust or if you'd like to discuss the migration further!

Thanks again


orangecoding:
Hey. Thanks so much for all the effort. Personally due to many reasons I'm not a big fan of typescript. (Whichever ehy Fredy is written in ts) 
Tbh I would rather rewrite it in go than to do it in ts ;-))


However your fork should not get lost. I could think of referring to your fork for those who like ts more than js. Question is if this is a one-of or if you want to take on maintaining it futurewise ;)

vncnt-dev:
Thanks for the honest response — totally understandable, and I appreciate you taking the time to consider it!

I’ll definitely continue maintaining the fork independently, but probably not as a strict 1:1 TypeScript mirror of the original. Most likely, I’ll selectively pull in some of the commits of this project where they make sense, while also making changes of my own. However, this means, that over time the two codebases diverge even more than they already have.

orangecoding:
That is fine & the nature of open source. I'm curious to see where your ideas will lead Fredy. If you have a good idea, you might want to give me a ping futurewise ;)

PR #142, opened 2025-07-22
Remove leading commas from listings in Telegram messages

alexanderroidl:
This MR includes:
* Remove comma prefixes in Telegram messages
* Icon alignment with text
* [Fixed an ESLint error](https://github.com/orangecoding/fredy/pull/142/files#diff-9ddbe1217b7598ed35e7a1d9d7d5f3f6eb10a4761008990d70a40f0da289caab)
* [Added a missing word to ImmoScout24 hint](https://github.com/orangecoding/fredy/pull/142/files#diff-9ddbe1217b7598ed35e7a1d9d7d5f3f6eb10a4761008990d70a40f0da289caab)

# Comma prefixes in Telegram messages

**The way Telegram message chunks were concatenated before lead to a leading comma for every subsequent listing after the first one found:**

telegram-comma

This is due to an array of strings being added to the message string directly:

https://github.com/orangecoding/fredy/blob/d89b0782378f52126855e9221389c48aedddf28e/lib/notification/adapter/telegram.js#L29-L34

I changed it to append to an array of strings, each acting as a paragraph, to be later concatenated using newlines.

# Icon alignment with text

As can be seen below, when setting all icon's `vertical-alignment` property to `middle`, they become perfectly aligned with the text, instead of being a little bit too high.

icon-alignment

### DIFF
--- a/lib/notification/adapter/telegram.js
+++ b/lib/notification/adapter/telegram.js
@@ -22,16 +22,18 @@ export const send = ({ serviceName, newListings, notificationConfig, jobKey }) =
   const { token, chatId } = notificationConfig.find((adapter) => adapter.id === config.id).fields;
   const job = getJob(jobKey);
   const jobName = job == null ? jobKey : job.name;
-  //we have to split messages into chunk, because otherwise messages are going to become too big and will fail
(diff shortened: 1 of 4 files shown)


orangecoding:
Looks good. Thanks for your contribution.

PR #144, opened 2025-07-22
Rename NPM startup scripts

alexanderroidl:
As a user of this application I find it quite hard to make a connection between
* `yarn prod` and the Backend
* `yarn start` and the Frontend

Both "prod" and "start" sound ambiguous and give no indication of what parts of the application they are going to start.

Therefore my suggestion is to
* Rename `yarn prod` to `yarn start:backend`
* Rename `yarn start` to `yarn start:frontend`

Also the Frontend startup via Yarn used to always be in Vite's Development mode (watch, restart on changes), while the Backend startup always was in Production mode (plain startup). 

I made it so both scripts offer a Development version using a `:dev` suffix (e.g. `yarn start:backend:dev`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "[F]ind [R]eal [E]states [d]amn eas[y].",
   "scripts": {
     "prepare": "husky",
-    "start:backend": "NODE_ENV=production node index.js",
+    "start:backend": "x-var NODE_ENV=production node index.js",
     "start:backend:dev": "nodemon --watch index.js --watch lib",
     "start:frontend": "vite -m production",
     "start:frontend:dev": "vite",
@@ -89,7 +89,8 @@
     "serve-static": "2.2.0",
     "slack": "11.0.2",
     "string-similarity": "^4.0.4",
-    "vite": "7.0.5"
+    "vite": "7.0.5",
+    "x-var": "^2.1.0"
   },
   "devDependencies": {
     "@babel/core": "7.27.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2675,6 +2675,11 @@ domutils@^3.0.1, domutils@^3.2.1, domutils@^3.2.2:
     domelementtype "^2.3.0"
     domhandler "^5.0.3"
 
+dotenv@^16.4.5:
+  version "16.6.1"
+  resolved "https://registry.yarnpkg.com/dotenv/-/dotenv-16.6.1.tgz#773f0e69527a8315c7285d5ee73c4459d20a8020"
+  integrity sha512-uBq4egWHTcTt33a72vpSG0z3HnPuIl6NqYcTrKEg2azoEyl2hpW0zqlxysq2pK9HlDIHyHyakeYaYnSAwd8bow==
+
 dunder-proto@^1.0.0, dunder-proto@^1.0.1:
   version "1.0.1"
   resolved "https://registry.yarnpkg.com/dunder-proto/-/dunder-proto-1.0.1.tgz#d7ae667e1dc83482f8b70fd0f6eefc50da30f58a"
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -23,9 +23,9 @@ If you want to try out _Fredy_, you can access the demo version [here](https://f
 - Make sure to use Node.js 20 or above
 - Run the following commands:
 ```ssh
-yarn (or npm install)
```
alexanderroidl, line 26: I would not advice for or offer support for another package manager here, but default to Yarn. 

Also `npm install` will create a `package-lock.json` file which is not in the `.gitignore`.

lib/services/immoscout/immoscout-web-translator.js
alexanderroidl, line 1: The filename did not match the file stated in `reverse-engineered-immoscout.md` and `translator` is the correct spelling.

package.json
```diff
@@ -4,10 +4,11 @@
   "description": "[F]ind [R]eal [E]states [d]amn eas[y].",
   "scripts": {
     "prepare": "husky",
-    "start": "node prod.js",
-    "dev": "yarn && rm -rf ./ui/public/* && vite",
```
alexanderroidl, line 8: We should assume the project dependencies to be installed before a script is executed:

```json
{
    "dev": "rm -rf ./ui/public/* && vite",
    "ui": "rm -rf ./ui/public/* && vite",
    "prod": "vite build --emptyOutDir",
}
```

Deletion of files inside `ui/public` can be handled via Vite's `emptyOutDir` option:

```json
{
    "dev": "vite --emptyOutDir",
    "ui": "vite --emptyOutDir",
    "prod": "vite build --emptyOutDir",
}
```

After setting the `emptyOutDir` option via `vite.config.js`:

```json
{
    "dev": "vite",
    "ui": "vite",
    "prod": "vite build",
}
```

Then after removing the duplicate script and renaming everything properly:

```json
{
    "start:frontend:dev": "vite",
    "build:frontend": "vite build"
}
```

package.json
```diff
@@ -4,10 +4,11 @@
...
-    "ui": "rm -rf ./ui/public/* && vite",
-    "prod": "yarn && vite build --emptyOutDir",
+    "start:backend": "NODE_ENV=production node index.js",
+    "start:backend:dev": "nodemon  --watch index.js --watch lib",
```
alexanderroidl, line 8: Nodemon uses the `package.json`'s `main` property by default (`index.js` in our case).

CONTRIBUTING.md
```diff
@@ -106,14 +106,14 @@ exports.config = {
...
+If you've written a new provider you are an awesome person. If you now write tests for it, you are even more awesome. And who doesn't want to be more awesome, right?
 
 #### Codestyle
-I'm using Eslint to maintain quote style and quality. Do not skip it...
+I'm using ESLint to maintain quote style and quality. Do not skip it...
 
-##### To do before merging:
+##### To-do before merging:
 
-- executed tests? (`pnpm test`)
```
alexanderroidl, line 116: See https://github.com/orangecoding/fredy/pull/144/files#r2223976502

ui/src/App.less
```diff
@@ -42,6 +42,6 @@ a:active {
   text-decoration: underline;
 }
 
-.semi-icon {
+.semi-icon:not(.semi-tabs-bar .semi-tabs-tab .semi-icon) {
```
alexanderroidl, line 45: This was causing trouble in the menu's tab bar (sorry, my mistake from earlier).

CONTRIBUTING.md
orangecoding: interesting. That was a leftover which I forgot to remove I assume

doc/reverse-engineered-immoscout.md
```diff
@@ -77,4 +77,4 @@ curl -H "User-Agent: ImmoScout24_1410_30_._" \
 
 
 ## Parameters 
-The parameters between web and mobile are very different which is why we have to translate them. Please see `immoscout-web-translator.js`.
\ No newline at end of file
+The parameters between web and mobile are very different which is why we have to translate them. Please see [/lib/services/immoscout/immoscout-web-translator.js](https://github.com/orangecoding/fredy/blob/master/lib/services/immoscout/immoscout-web-translator.js).
```
orangecoding, line 80: Please revery moving the doc. It is linked in numerous blogs and if we move it, we break all those links. Thanks

orangecoding (changes requested):
I important comment.

doc/reverse-engineered-immoscout.md
alexanderroidl: Moved it back!

alexanderroidl:
@orangecoding I also added a status badge for the workflow I added to the README with my latest commit.

orangecoding:
@alexanderroidl there is 1 more conflict. When this is resolved, it should be ok to get merged.

orangecoding (changes requested):
PLZ resolve that one conflict, than we can merge it

PR #148, opened 2025-07-26
Replace runtime imports with static imports

alexanderroidl:
Currently there are dynamic imports at runtime requiring programmatically reading the file system and performing asynchronous operations, which can be avoided by indexing all desired files in an `index.js` and statically importing that file instead.

### DIFF
--- a/lib/notification/adapter/index.js
+++ b/lib/notification/adapter/index.js
@@ -1,0 +1,12 @@
+import * as apprise from './apprise.js';
+import * as console from './console.js';
+import * as mailJet from './mailJet.js';
+import * as mattermost from './mattermost.js';
+import * as ntfy from './ntfy.js';
+import * as pushover from './pushover.js';
+import * as sendGrid from './sendGrid.js';
+import * as slack from './slack.js';
+import * as sqlite from './sqlite.js';
+import * as telegram from './telegram.js';
+
+export default [apprise, console, mailJet, mattermost, ntfy, pushover, sendGrid, slack, sqlite, telegram];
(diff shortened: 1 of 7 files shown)


alexanderroidl:
@orangecoding Unrelated to this but apparently ESLint warnings now show below the GitHub's change overview, that's kind of cool even if unintended 👍

Screenshot 2025-07-26 115855

alexanderroidl:
> > @orangecoding Unrelated to this but apparently due to the new workflow ESLint warnings now show up below GitHub's change overview 😊 Kind of cool even if unintended
> 
> Yep, got a notification about that a couple of weeks back.

I think it's new actually, on my MR four days ago it wasn't there: https://github.com/orangecoding/fredy/pull/142/files

lib/notification/adapter/index.js
```diff
@@ -0,0 +1,12 @@
+import * as apprise from './apprise.js';
```
orangecoding, line 1: I know I know, this makes things more performant.. BUT, that comes with a price. The reason I did it the way it is today is because ease of use. You can simply add a new file, throw it into the folder and you can be sure it will be read and it is available. You do not need to register it somewhere. This ease of use of course comes with the price of a little bit of performance, but since Fredy is not rly a high performant app, I'm willing to pay the price.

What's your take on this?

orangecoding (changes requested):

lib/notification/adapter/index.js
alexanderroidl: Regardless of performance I don't think our case justifies a custom loading implementation. I would expect anybody implementing a provider capable of adding to the index too and it not being too much of an extra effort

But other than that, I would like for providers and notif. adapters to use their own directory and multiple files. That would mean we would have to further complicate our logic to support this :/ What do you think?

orangecoding:
@alexanderroidl I'd close this pr as I don't see any activity. Thanks a lot tho

PR #149, opened 2025-07-26
Lodash methods and helper functions

alexanderroidl:
This MR aims to refactor by

- Switching from our `nullOrEmpty` helper functions to check for [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) or [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) values natively.
- Switching some methods on our end for `Lodash` methods as it's a dependency of ours anyways

### DIFF
--- a/lib/api/routes/userRoute.js
+++ b/lib/api/routes/userRoute.js
@@ -10,7 +10,6 @@ function checkIfAnyAdminAfterRemovingUser(userIdToBeRemoved, allUser) {
 function checkIfUserToBeRemovedIsLoggedIn(userIdToBeRemoved, req) {
   return req.session.currentUser === userIdToBeRemoved;
 }
-const nullOrEmpty = (str) => str == null || str.length === 0;
 userRouter.get('/', async (req, res) => {
   res.body = userStorage.getUsers(false);
   res.send();
@@ -42,7 +41,6 @@ userRouter.delete('/', async (req, res) => {
   res.send();
 });
 userRouter.post('/', async (req, res) => {
-
   if(config.demoMode){
       res.send(new Error('In demo mode, it is not allowed to change or add user.'));
       return;
@@ -53,7 +51,7 @@ userRouter.post('/', async (req, res) => {
     res.send(new Error('Passwords does not match'));
     return;
   }
-  if (nullOrEmpty(username) || nullOrEmpty(password) || nullOrEmpty(password2)) {
+  if (!username || !password || !password2) {
     res.send(new Error('Username and password are mandatory.'));
     return;
   }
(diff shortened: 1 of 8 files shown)


lib/api/routes/userRoute.js
```diff
@@ -53,7 +51,7 @@ userRouter.post('/', async (req, res) => {
     res.send(new Error('Passwords does not match'));
     return;
   }
-  if (nullOrEmpty(username) || nullOrEmpty(password) || nullOrEmpty(password2)) {
+  if (!username || !password || !password2) {
```
orangecoding, line 54: See the previous pr. your check is not the same as mine.
```
let a = 0;
nullOrEmpty(a); //false
!a; //true
```

orangecoding: in my case I want to explicitly check for null or empty

lib/notification/adapter/telegram.js
```diff
@@ -1,6 +1,8 @@
 import { markdown2Html } from '../../services/markdown.js';
 import { getJob } from '../../services/storage/jobStorage.js';
 import fetch from 'node-fetch';
+import lodash from 'lodash';
```
orangecoding, line 4: import {chunk} from 'lodash';

lib/utils.js
```diff
@@ -29,7 +25,7 @@ function timeStringToMs(timeString, now) {
 
 function duringWorkingHoursOrNotSet(config, now) {
     const {workingHours} = config;
-    if (workingHours == null || nullOrEmpty(workingHours.from) || nullOrEmpty(workingHours.to)) {
+    if (!workingHours?.from || !workingHours?.to) {
```
orangecoding, line 28: I'd consider checking for null or empty by doing this truthy/falsy check relatively unsafe for reason mentioned above. WDYT?

orangecoding (changes requested):

lib/api/routes/userRoute.js
alexanderroidl: Sounds good, I changed it back to the same logic but slimer:

```js
const nullOrEmpty = (str) => str == null || str.length === 0;
```

is the same as
```js
const nullOrEmpty = (str) => Boolean(str?.length);
```

as the `if`-block will convert everything inside it's brackets to a boolean value for checking, we can inline it:

```js
  if (!username?.length || !password?.length || !password2?.length) {
```

Let me know if that works for you.

orangecoding (changes requested):
I added 1 comment. I think your "nullOrEmpty" check does not work as intended. Also there are now a couple of conflicts. Would you mind resolving them? :)

orangecoding (changes requested):
This pr results in Test failures. Tbh, I'm loving what you have contributed so far, but I'm not convinced about this one...

Issue #178, opened 2025-09-18
[Bug]: 12.0.0 fails to run in Docker


CommanderTvis:
### Bug Description

The latest release fails to run in Docker.

### Steps to Reproduce

```shell
git clone ...
git checkout 12.0.0
docker compose up --build
```

### Expected Behavior

The website starts successfully.

### Actual Behavior

After building, the process fails to start with
```
Attaching to fredy
fredy  | 2025-09-18T15:05:16: PM2 log: Launching in no daemon mode
fredy  | 2025-09-18T15:05:16: PM2 log: Launching in no daemon mode
fredy  | 2025-09-18T15:05:16: PM2 log: App [index:0] starting in -fork mode-
fredy  | 2025-09-18T15:05:16: PM2 log: App [index:0] online
fredy  | 2025-09-18T15:05:16: PM2 log: App [index:0] starting in -fork mode-
fredy  | 2025-09-18T15:05:16: PM2 log: App [index:0] online
fredy  | Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/lib/services/storage/SqliteConnection.js' imported from /db/migrations/migrate.js
fredy  |     at finalizeResolution (node:internal/modules/esm/resolve:274:11)
fredy  | Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/lib/services/storage/SqliteConnection.js' imported from /db/migrations/migrate.js
fredy  |     at finalizeResolution (node:internal/modules/esm/resolve:274:11)
fredy  |     at moduleResolve (node:internal/modules/esm/resolve:859:10)
fredy  |     at defaultResolve (node:internal/modules/esm/resolve:983:11)
fredy  |     at #cachedDefaultResolve (node:internal/modules/esm/loader:717:20)
fredy  |     at ModuleLoader.resolve (node:internal/modules/esm/loader:694:38)
fredy  |     at ModuleLoader.getModuleJobForImport (node:internal/modules/esm/loader:308:38)
fredy  |     at ModuleJob._link (node:internal/modules/esm/module_job:183:49)
fredy  |     at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
fredy  |   code: 'ERR_MODULE_NOT_FOUND',
fredy  |   url: 'file:///lib/services/storage/SqliteConnection.js'
fredy  | }
fredy  |     at moduleResolve (node:internal/modules/esm/resolve:859:10)
[...]

PR #179, opened 2025-09-18
Add health check for Docker container

CommanderTvis:
- Introduced `HealthCheck` in `docker-compose.yml` to monitor container status.
- Added a test step to validate container's health using Docker Compose in the GitHub workflow.
- Updated `Dockerfile` to include `curl` for health check commands.


CommanderTvis:
The suggested workflow more or less protects the repository from deploying a non-running Docker container, which can help to prevent errors like #178 in future.

PR #194, opened 2025-09-25
Fix and improve immowelt/immonet provider

gorgobacka:
- Parse description
- Use correct id for image


orangecoding (changes requested):
Thanks for this, however, there seem to be test failures. Can you run tests and make sure to fix this? Thanks :)

Issue #205, opened 2025-09-28
[Bug]: Telegram silently breaks for response code 400 (Bad request)


gorgobacka:
### Bug Description

While playing around with the new Sparkasse provider, I detected that not all listings are send to Telegram.
After some analysis, I found the issue. Telegram does not like all image urls.

E.g. This image is included in a listing in the Sparkasse provider: https://api.immobilie1.de/estates/31829381/image/278406671/expose-estate-gallery-main-desktop-avif.avif

The image load just fine in a web browser but Telegram does not like it.

### Steps to Reproduce

Modify the code to send a message to Telegram with the image URL: https://api.immobilie1.de/estates/31829381/image/278406671/expose-estate-gallery-main-desktop-avif.avif

### Expected Behavior

The error is caught and a message is send without an image (`sendMessage`).

### Actual Behavior

No message to Telegram is send and no log message appears. Therefore, the user might miss listings, if he does not check the Fredy UI, carefully.

### Screenshots / Logs

Status Code: 400
Text: `{"ok":false,"error_code":400,"description":"Bad Request: wrong type of the web page content"}`

### Environment

Ubuntu

### Additional Context

I don't know, if other providers have this problem, too.

gorgobacka:
Thanks a lot for improving my code! Sorry for my late reply but I'm on vacation currently.

PR #211, opened 2025-09-29
Check Telegram response (#205)

gorgobacka:
The following changes are included in this PR:
- check the response from Telegram, if the request was successful
  - if not throw an error and print a log message
  - retry with `sendMessage` as a fallback

- add more debug messages including the name of the provider (I find this quite useful to understand, what happened)
   - number of detected listings
   - number of new listings
   - if no new listings were detected

It's a while since I worked with JS/TS and unfortunately, I don't know how to provide a test for this scenario.
I tested this manually with an incorrect image url and it worked fine.

Please let me know, if it fits your scope of your project and what can be improved.

### DIFF
--- a/lib/FredyRuntime.js
+++ b/lib/FredyRuntime.js
@@ -77,6 +77,7 @@ class FredyRuntime {
   }
 
   _findNew(listings) {
+    logger.debug(`Checking ${listings.length} listings for new entries (Provider: '${this._providerId}')`);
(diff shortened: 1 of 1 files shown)


gorgobacka:
Sorry, I do not get it. Can you please rephrase it?

The problem is that the response by Telegram is not checked, if it was successful or not and nothing is shown in the logs, if the message was not send.

I will think about a different solution.

What would be your idea to solve it?

PR #214, opened 2025-09-30
Use non-root user in Dockerfile

### LINKED ISSUE #207: [Feature]: Make app run as non-root in container
### Related Problem

It is a security best practice to make your service run as non-root inside the container: https://docs.docker.com/build/building/best-practices/#user

Especially when it is executed in a rootful way this increases the security.

### Proposed Feature

Adjust the Dockerfile to drop permissions when not necessary (anymore).

### Alternatives Considered

_No response_

### Benefits

See https://stackoverflow.com/a/68190847/5008962

### Additional Context

_No response_

CommanderTvis:
closes #207

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,11 +8,16 @@ RUN apt-get update \
   && apt-get install -y --no-install-recommends chromium curl \
   && rm -rf /var/lib/apt/lists/*
 
+# Use predefined node user (UID 1000, GID 1000)
+# Set home directory ownership for node user
+RUN usermod -d /fredy node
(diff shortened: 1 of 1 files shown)


orangecoding:
Merged. Thanks.

orangecoding:
@CommanderTvis i merged this, but the build fails!
Can you check please? Some error with accessing the db files. Otherwise I have to roll it back

https://github.com/orangecoding/fredy/actions/runs/18171208969/job/51726122064#step:8:1

PR #219, opened 2025-10-24
added additional pictures for listings

dosr1:
new feature: added additional pictures for listings (working for immoscout provider first)

### DIFF
--- a/lib/provider/immoscout.js
+++ b/lib/provider/immoscout.js
@@ -59,15 +59,59 @@ async function getListings(url) {
     logger.error('Error fetching data from ImmoScout Mobile API:', response.statusText);
     return [];
   }
-
   const responseBody = await response.json();
-  return responseBody.resultListItems
+  // helper: convert media entries (objects or strings) to URL strings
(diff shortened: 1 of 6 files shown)


lib/provider/immoscout.js
```diff
@@ -76,8 +120,87 @@ async function getListings(url) {
         link: `${metaInformation.baseUrl}expose/${item.id}`,
         address: item.address?.line,
         image,
+        additionalpictures: extractedAdditionalPictures,
+        detailedinfo: null,
       };
     });
+
+  // Fetch detailed info for all items in parallel and attach results
+  await Promise.all(
```
orangecoding: I was thinking about that and even playing around with this, but I strongly suggest _not_ to do this, because this will spawn way too many requests. It _might_ work for Immoscout, however it will not work for others. Immonet/Immowelt, all of which work the "classic" way will detect these parallel calls as bots, thus subsequent calls will fail.

orangecoding (changes requested):

PR #231, opened 2025-11-13
feat: add http adapter

efekrskl:
Adds a generic HTTP adapter for making a `POST` request at a target URL after new listings are found.

My use case would be to have an endpoint, that I can make a request to when new listings are found. This way the user could create long-running servers, or serverless functions to do additional operations.

### DIFF
--- a/lib/notification/adapter/http.md
+++ b/lib/notification/adapter/http.md
@@ -5,6 +5,7 @@ You can leverage this adapter to integrate with various webhooks or APIs that ac
 Functions, a Node.js server, etc.)
 
 HTTP adapter supports a `authToken` field, which can be used to include an authorization token in the request headers.
+Your token would be included as a Bearer token in the `Authorization` header, which is a common method for securing API requests.
 
 Request Details:
 <details>


lib/notification/adapter/http.md
```diff
@@ -0,0 +1,31 @@
+### HTTP Adapter
+
+This is a generic adapter for sending notifications via HTTP requests.
```
orangecoding: You're not mentioning the auth token here. I'd make it optional, but describe how to use it in any case. Note, not all users are pro's ;)

lib/notification/adapter/http.js
```diff
@@ -0,0 +1,57 @@
...
+  readme: markdown2Html('lib/notification/adapter/http.md'),
+  description: 'Fredy will send a generic HTTP POST request.',
+  fields: {
+    endpointUrl: {
+      description: "Your application's endpoint URL.",
+      label: 'Endpoint URL',
+      type: 'text',
+    },
+    authToken: {
+      description: "Your application's auth token, if required by your endpoint.",
```
orangecoding: I'll merge it this time, but in the future, please make sure to install the pre-commit hooks. If you do, it would have told you that double quotes are not allowed. I'll fix this later

efekrskl: This string has a single quote inside, therefore it doesn't raise any issues. This is how similar usages are handled (e.g. `lib/notification/adapter/telegram.js:122`)

Issue #245, opened 2025-12-12
[Bug]: Unraid - no access to webui


sheddixx:
### Bug Description

on version 15.0.1 everything works fine
after upgrading to 16.0.0, I cant access the webui.
container doesnt crash, it is still running. log closes automatically.
only thing i changed was the default port., but on 15.0.1 ist still working.

### Steps to Reproduce

change port from default
upgrade to 16.0.0

### Expected Behavior

accessing the webui

### Actual Behavior

cant access the webui

### Screenshots / Logs

<img width="1016" height="654" alt="Image" src="https://github.com/user-attachments/assets/7c70e3e2-3e0b-449c-a44a-df9f08218833" />

### Environment

unraid 7.2.2

Issue #244, opened 2025-12-11
[Bug]: Immoscout translator crash application when type is unknown


lorem-ipsum-dolor-sit:
### Bug Description

The Docker ist crashing, after finding a townhouse search.

### Steps to Reproduce

Use as a townhouse search: 

https://www.immobilienscout24.de/Suche/de/hamburg/townhouse

### Expected Behavior

Not crashing :)

### Actual Behavior

Crashing the docker.

### Screenshots / Logs
```
Node.js v22.21.1
[2025-12-11 20:34:03] INFO: Started Fredy successfully. Ui can be accessed via http://localhost:9998
[2025-12-11 20:34:03] DEBUG: Started API service on port 9998
file:///fredy/lib/services/immoscout/immoscout-web-translator.js:155
      throw new Error(`Real estate type not found: ${realTypeKey}`);
            ^
Error: Real estate type not found: townhouse
    at convertWebToMobile (file:///fredy/lib/services/immoscout/immoscout-web-translator.js:155:13)
    at Module.init (file:///fredy/lib/provider/immoscout.js:138:16)
    at file:///fredy/index.js:91:31
    at Array.forEach (<anonymous>)
    at file:///fredy/index.js:89:14
    at Array.forEach (<anonymous>)
    at execute (file:///fredy/index.js:86:10)
    at file:///fredy/index.js:109:1
```
### Environment

v16.0.0

PR #246, opened 2025-12-12
Fix startup crash by catching and logging errors instead of exiting

lorem-ipsum-dolor-sit:
Relates to #244 and #245.

The application previously exited immediately when `matchedProvider.init(prov, job.blacklist)` threw an exception during startup, causing unreliable startups in certain environments. This change wraps the init logic in a try/catch that logs the error and allows the app to continue running.


orangecoding (approved):

sheddixx:
Thank you very much, great job. 16.0.1 workong now 👍🏼

PR #273, opened 2026-02-22
feat(): map area filter

strech345:
- add `spatial_filter` to db from type `geoJson` or `null`
- add areaFilter Map to a job config
- create `Map` component
- create `MapDrawingExtension` 
- use `spatial_filter` to filter listings

### DIFF
--- a/test/provider/immobilienDe.test.js
+++ b/test/provider/immobilienDe.test.js
@@ -14,7 +14,7 @@ describe('#immobilien.de testsuite()', () => {
   it('should test immobilien.de provider', async () => {
     const Fredy = await mockFredy();
     return await new Promise((resolve) => {
-      const fredy = new Fredy(provider.config, null, provider.metaInformation.id, 'test1', similarityCache);
+      const fredy = new Fredy(provider.config, null, null, provider.metaInformation.id, 'test1', similarityCache);
       fredy.execute().then((listing) => {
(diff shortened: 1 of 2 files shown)


ui/src/views/jobs/mutation/components/areaFilter/AreaFilter.less
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2026 by Christian Kellner.
+ * Licensed under Apache-2.0 with Commons Clause and Attribution/Naming Clause
+ */
+
+.areaFilter-container {
+  height: 500px;
```
orangecoding: plz use rem instead of px.

Also please use less style. E.g.

```
.areaFilter {
  &__container{
   ... 
 }
}
```

strech345:
@orangecoding i would also like to add spec filters: for minSquareMeters, minRools and maxPrice. Would add it in same way like the areaFilter if this ok for you.

orangecoding:
@strech345 Thanks,

I'll merge your changes and fix the test issues in my branch. Also I found an issue, not introduced by you, but by a logical issue. We filter out listings to not send then via notifications if they're out of the geo area. also if they're a similar entry to something else, however What I still do is storing it and not tagging it as duplicate/filtered. This way it still appears in the map.

I'm going to fix this before releasing your change.

PR #276, opened 2026-03-08
Feature/spec filter

strech345:
listings can be filtered by specs
- specs are: price, size, rooms
- by fiilters: maxPrice, minSize, minRooms
- will be filtered out for notifications
- will be filtered out on overview
- for now its only specFilters, but also blacklist and areaFilter should be added here.
- can be disabled on overview (in filter section)
- add rooms into db because its next to price and size one of the three importen specs
- show rooms in overview and detail
- update kleinanzeigen config
- add rooms for now only on kleinanzeigen but on the other it should also be added


strech345:
> So should I review it now or is this still a draft?

Would be nice if you can read my comments here and give me some Feedback. Also you can Start reviewing it. I would only extend it like explained.

strech345:
> @strech345 You probably want to merge Master into your branch as I have just pushed a pretty big change. (There's a new migration as well, so you must increase the number of yours) See https://github.com/orangecoding/fredy/releases/tag/20.0.0

whats about migration `11.add-spatial-filter`? that was on master merged before. Now we have two 11s (11.mcp-tokens).

orangecoding:
@claude review

orangecoding:
@strech345 What's the status here?

strech345:
> @strech345 What's the status here?

i did your requests and merged newest version. get many conflicts. Hope everything is now fine. Please check.

orangecoding:
Thanks for all the work you put into this! @strech345

PR #289, opened 2026-03-30
Feature/Kleinanzeigen addresses

realDayaa:
It was bugging me that Kleinanzeigen listings only provided the postal code and not the full address.
As a result, they didn't show up on the map correctly.

This PR provides a toggle similar to Immoscout Details that allows the scraping of individual listings so as to provide a full address and description.


ui/src/views/generalSettings/GeneralSettings.jsx
```diff
@@ -465,6 +466,32 @@ const GeneralSettings = function GeneralSettings() {
                   </div>
                 </SegmentPart>
 
+                <SegmentPart
```
orangecoding: Why do you put this both in the General Settings AND the user Settings? I'd vote for just putting it into the UserSettings

orangecoding: How do you want to go about this. Do you want me to merge your pr, and then build up on top of it?
Would that be ok with you?

PR #299, opened 2026-04-15
Add toggle for sending telegram message as plain text message

adrianbartnik:
Hi,

first, thanks for creating this service and making it open source - really useful 🙏 

This PR adds a toggle to the Telegram Notification Adapter to allow sending the messages in plain text instead of HTML. I personally prefer this, because I don't have to click again to open the link, see screenshot below. Also, the images are still rendered, because telegram generates a preview of the listing based on the link on the client.

SCR-20260415-tuzz

For transparency, I have used Claude Code to add those changes, because I am not that familiar with JS.

Rendered toggle configuration in the adapter configuration:
SCR-20260415-tqti

<table>
<tr>
 <td>HTML Version:
 <td>Plain Text Version
<tr>
 <td>SCR-20260415-trqb
 <td>SCR-20260415-trnl
</table>

### DIFF
--- a/lib/notification/adapter/telegram.js
+++ b/lib/notification/adapter/telegram.js
@@ -106,6 +106,32 @@ function buildText(jobName, serviceName, o) {
 }
 
 /**
+ * Build a plain text Telegram photo caption (max 4096 characters).
+ * @param {string} jobName
+ * @param {string} serviceName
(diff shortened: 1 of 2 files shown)


orangecoding (approved):
LGTM

PR #300, opened 2026-04-17
change average price to median price on the dashboard

bytedream:
Shows the median price instead of the average on the dashboard. This is more resilient to high priced or irregular listings.
For example, I've encountered a listing from someone who was searching, not offering, a flat. They set an imaginary price of 123456€ which impacted the average price significantly, the average price on my dashboard is now ~7k, whereas the median price is around 500€, which is far more accurate. I could manually remove it, but the problem that other flats which are too high priced, that somehow found their way into my search query, still persists.

### DIFF
--- a/lib/services/storage/listingsStorage.js
+++ b/lib/services/storage/listingsStorage.js
@@ -29,33 +29,46 @@ export const getKnownListingHashesForJobAndProvider = (jobId, providerId) => {
  * Compute KPI aggregates for a given set of job IDs from the listings table.
  *
  * - numberOfActiveListings: count of listings where is_active = 1
- * - avgPriceOfListings: average of numeric price, rounded to nearest integer
+ * - medianPriceOfListings: median of numeric price, rounded to nearest integer
  *
  * When no jobIds are provided, returns zeros.
  *
  * @param {string[]} jobIds
- * @returns {{ numberOfActiveListings: number, avgPriceOfListings: number }}
+ * @returns {{ numberOfActiveListings: number, medianPriceOfListings: number }}
  */
 export const getListingsKpisForJobIds = (jobIds = []) => {
   if (!Array.isArray(jobIds) || jobIds.length === 0) {
-    return { numberOfActiveListings: 0, avgPriceOfListings: 0 };
+    return { numberOfActiveListings: 0, medianPriceOfListings: 0 };
   }
 
   const placeholders = jobIds.map(() => '?').join(',');
-  const row =
-    SqliteConnection.query(
-      `SELECT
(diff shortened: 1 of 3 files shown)


bytedream:
The calculation of the median price is done in js. In theory sqlite supports the `median(x)` selector, but `better-sqlite3` doesn't build sqlite with the needed flags for it, so it isn't available. `better-sqlite3` supports building sqlite with custom build flags, which could be used to enable `median(x)` support, but idk if it's worth to recompile sqlite just for this one feature.

lib/services/storage/listingsStorage.js
```diff
@@ -29,33 +29,46 @@ export const getKnownListingHashesForJobAndProvider = (jobId, providerId) => {
...
-    SqliteConnection.query(
-      `SELECT
-          SUM(CASE WHEN is_active = 1 THEN 1 ELSE 0 END) AS activeCount,
-          AVG(price) AS avgPrice
-       FROM listings
-       WHERE job_id IN (${placeholders})
-         AND manually_deleted = 0`,
-      jobIds,
-    )[0] || {};
+  const rows = SqliteConnection.query(
```
orangecoding, line 45: Mhm. maybe like so? A little more efficient..


```suggestion
const rows = SqliteConnection.query(
  `SELECT
     SUM(CASE WHEN is_active = 1 THEN 1 ELSE 0 END) AS active_count,
     price
   FROM listings
   WHERE job_id IN (${placeholders})
     AND manually_deleted = 0`,
  jobIds,
);

const activeCount = rows[0]?.active_count ?? 0;

const prices = rows
  .map(r => r.price)
  .filter(p => p !== null)
  .sort((a, b) => a - b);

let medianPrice = 0;
if (prices.length > 0) {
  const mid = Math.floor(prices.length / 2);
  medianPrice = prices.length % 2 !== 0 ? prices[mid] : (prices[mid - 1] + prices[mid]) / 2;
}

return {
  numberOfActiveListings: activeCount,
  medianPriceOfListings: medianPrice,
};
```

orangecoding (changes requested):

lib/services/storage/listingsStorage.js
bytedream: That's indeed better, ty for the suggestion

bytedream: Argh mb, this doesn't work. Because of the `sum` selector, the sql query only returns one row

bytedream: Okay I think I fixed it now, see my changes

PR #304, opened 2026-05-06
feat: Add delete button to listing detail view

datenwurm:
Adds a delete button to the listing detail view, allowing users to delete a listing without navigating back to the overview.

**Behavior:**
Reuses the soft-/hard-delete dialog from the listings overview view to keep deletion behavior consistent.
After successful deletion, the user is redirected to the listings overview.

**Motivation:**
Previously, deleting a listing from the detail view required going back to the overview, locating the listing again (search and scroll), and clicking delete there. Users can now delete directly from the detail view.


orangecoding:
Thanks!

PR #305, opened 2026-05-06
feat: Add grid/table view toggle to listings overview

datenwurm:
Adds two buttons to the listings overview that let users switch between the existing grid view and a new table view.

The selected view is persisted in user settings, so it remains active across sessions.
The table view exposes the same listing data as the grid, in a denser, row-based layout.

image


orangecoding:
@datenwurm looks good to me. Would you mind doing the same for the jobs overview? :D

orangecoding:
> @datenwurm looks good to me. Would you mind doing the same for the jobs overview? :D

Done it myself ;)

https://github.com/orangecoding/fredy/releases/tag/21.3.0

PR #308, opened 2026-05-14
fix: kleinanzige listing might have a different structure 

strech345:
when a listing was created with no image the preview will have a different structure, also when the image was added later.


orangecoding:
@strech345 did you run the tests (non offline ones)?

strech345:
@orangecoding you mean `yarn test`?

i got a error which i think not relay on my change
```
 test/provider/immoswp.test.js > #immoswp testsuite() > should test immoswp provider 60040ms
   → Test timed out in 60000ms.
If this is a long-running test, pass a timeout value as the last argument or configure it globally with "testTimeout".
```

not shure if other tests are missing because of this error.

Issue #310, opened 2026-05-30
[Bug]: Area filter does not remove out-of-area listings


AbdallahZerfaoui:
### Bug Description

Hi maintainers, thanks for the great project.

I found what looks like a bug in the filtering pipeline: listings that should be removed by area/spec/similarity filters still appear in results.

### What I observed

When using a provider URL with a large radius (e.g. Kleinanzeigen 20km) and defining a smaller Area Filter in Fredy, listings outside my drawn polygon are still visible.

### Steps to Reproduce

1. Open Kleinanzeigen in browser.
2. Search for apartments in **Heilbronn** and set radius to **20 km**.
3. Copy the resulting search URL.
4. In Fredy, create a new job:
   - Provider: `Kleinanzeigen`
   - URL: paste the copied Kleinanzeigen URL
   - Keep defaults for other filters.
5. In the same job, open **Area Filter** and draw a small polygon only around **Heilbronn city center** (much smaller than 20 km radius).
6. Save the job.
7. Run the job once (`Run` in Jobs view).
8. Open Map view for that job.

### Expected Behavior

Listings filtered out by area/spec/similarity should be hidden/marked deleted reliably.

### Actual Behavior

Filtered listings can remain visible because cleanup targets the wrong identifier.

### Environment

Linux homeserver 6.8.0-90-generic #91-Ubuntu SMP PREEMPT_DYNAMIC Tue Nov 18 14:14:30 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux

---

If this diagnosis looks correct, I’d be happy to open a PR with a fix.  
Would you like me to proceed?

orangecoding:
Thanks! What do you think about this?

https://github.com/orangecoding/fredy/pull/311/changes

AbdallahZerfaoui:
I have tested fixing-listing-filtering branch, and it has generated a new bug.

**New Bug Description**:

1. Spatial filtering is applied only to `newListings`, not to the full current result set.
2. “Known listing” detection still includes hashes of `manually_deleted` rows.
3. Deleted hashes are never revived (`ON CONFLICT ... DO NOTHING`), so once filtered out, they can stay gone forever.
4. Hash-based delete is global (`WHERE hash IN (...)`) and not scoped by job/provider, so one job’s filtering can affect others.

orangecoding:
1. That is intended. It should only be applied to the list of the current execution
2. that's also intended. Why would you ewant to have duplicates?
[...]

PR #311, opened 2026-05-30
fixing filtering of lists

orangecoding:
Problem...
In storeListings, each listing row is saved with two distinct identifiers:
  - id = a fresh nanoid() (the DB primary key)
  - hash = item.id (the provider-assigned listing ID, e.g. "12345" from Kleinanzeigen)
  
  After _save, the listing objects in memory still carry the provider ID in listing.id. The three post-save filter methods (_filterBySimilarListings, _filterBySpecs, _filterByArea) all collected those values and passed them to
  deleteListingsById, which ran:

  UPDATE listings SET manually_deleted = 1 WHERE id IN (?)

  The id column holds nanoids, so the provider IDs never matched, the soft-delete was a no-op and filtered listings remained visible.

### DIFF
--- a/lib/FredyPipelineExecutioner.js
+++ b/lib/FredyPipelineExecutioner.js
@@ -7,7 +7,7 @@ import { NoNewListingsWarning } from './errors.js';
 import {
   storeListings,
   getKnownListingHashesForJobAndProvider,
-  deleteListingsById,
+  deleteListingsByHash,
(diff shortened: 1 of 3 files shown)


AbdallahZerfaoui:
I have tested fixing-listing-filtering branch, and it has generated a new [bug](https://github.com/orangecoding/fredy/issues/310#issuecomment-4586028071).

PR #325, opened 2026-06-05
added redirect after login

AdriDevelopsThings:
The old behaviour was: You open a page without being authorized, you are getting redirected to /login that redirects you after a successful authentication to /dashboard. This is really annoying if you want to open listenings directly from your notification adapter for example. This pull request introduces a method to redirect you back to the original page you opened after the authentication process by adding the navigation of the opened page as state to the navigation to /login. The login component than unpacks the state that contains the old navigation and redirects the user back to path from the original navigation. The path /dashboard is used as a fallback if no navigation in the state is present.

Freel free to write me if you have any questions or ideas to change here.


orangecoding (approved):
Looks good. Thanks!

PR #327, opened 2026-06-08
fix(dashboard): persist lastRun so scheduling KPIs stay in sync

### LINKED ISSUE #324: [Bug]: Last and Next search in dashboard are wrong
### Bug Description

The information for last and next search are not filled. However the jobs run and notifications were send.

Image

### Steps to Reproduce

Create a job. Let that job run.

### Expected Behavior

Display scheduling information

### Actual Behavior

Dont display scheduling info

### Screenshots / Logs

Image

Image

Image

### Environment

Freddy v22.4.0 running with docker compose on Ubuntu 24.04.4 LTS (GNU/Linux 6.8.0-124-generic x86_64)

### Additional Context

_No response_

RaminGe:
lastRun was only updated on the startup settings object while the dashboard reads a rebuilt cache after upsertSettings invalidates it.

Fixes: #324

### DIFF
--- a/lib/services/jobs/jobExecutionService.js
+++ b/lib/services/jobs/jobExecutionService.js
@@ -14,7 +14,7 @@ import * as similarityCache from '../similarity-check/similarityCache.js';
 import { isRunning, markFinished, markRunning } from './run-state.js';
 import { sendToUsers } from '../sse/sse-broker.js';
 import * as puppeteerExtractor from '../extractor/puppeteerExtractor.js';
-import { getSettings } from '../storage/settingsStorage.js';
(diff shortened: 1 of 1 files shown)


orangecoding:
Hey @RaminGe 

there was already a pr in the past which I closed. See https://github.com/orangecoding/fredy/pull/303

It's correct that this is not stored in the settings, simply because it is not a setting. I have not seen a value yet in storing this information in the db, but if I would, I certainly would store it as part of the job :)

I'll keep your change in mind and will do it at a later stage. Thanks for your contribution though!

PR #332, opened 2026-06-10
fix(immoscout): map exclusioncriteria swapflat to mobile API value swap_flat

### LINKED ISSUE #331: Immoscout searches with exclusioncriteria=swapflat silently return 0 results
## What happens

Any Immoscout search URL that contains `exclusioncriteria=swapflat` (the web UI's "Tauschwohnungen ausschließen" filter) never returns a single listing. No error is logged — the job just quietly finds nothing, which makes this really nasty to track down.

## Why

`convertWebToMobile()` in `lib/services/immoscout/immoscout-web-translator.js` passes `exclusioncriteria` through verbatim. The mobile API doesn't know the web value `swapflat` — it expects `swap_flat`. And instead of ignoring an unknown value, the API responds with `totalResults: 0` for the entire search.

## Repro

Same shape search (Munich rentals, ≤1500 €, 1.5+ rooms) against `https://api.mobile.immobilienscout24.de/search/list`, only the exclusion value changed:

| `exclusioncriteria` | totalResults |
|---|---|
| _(none)_ | 23 |
| `swapflat` (what the translator sends today) | **0** |
| `swap_flat` | 23 |
| `projectlisting,swapflat` | **0** |
| `projectlisting,swap_flat` | 23 |

Counts were stable across repeated calls. Note `projectlisting` is fine as-is — both APIs share that value, it's only `swapflat` that differs.

## Fix

Map `swapflat` → `swap_flat` during web→mobile conversion, leave every other value untouched. PR incoming.

michiosw:
Fixes #331

## Problem

The web UI encodes "exclude swap flats" as `exclusioncriteria=swapflat`, but the mobile API only understands `swap_flat`. The translator passes the value through verbatim, and the API doesn't ignore unknown values — it silently returns `totalResults: 0` for the whole search. So every saved search with that filter set finds nothing, with no error anywhere.

## Fix

Added an `EXCLUSION_CRITERIA_MAP` (analogous to the existing `EQUIPMENT_MAP`) that maps `swapflat` → `swap_flat` during web→mobile conversion. All other values pass through unchanged — `projectlisting`, for example, is identical on both APIs and keeps working (verified against the live API: it actually filters results when sent as-is, and is ignored as `project_listing`).

## Verification

Live mobile API, identical shape search (Munich rentals, ≤1500 €, 1.5+ rooms), stable across repeated calls:

| `exclusioncriteria` sent | totalResults |
|---|---|
| _(none)_ | 23 |
| `swapflat` (before this PR) | **0** |
| `swap_flat` (after this PR) | 23 |
| `projectlisting,swapflat` (before) | **0** |
| `projectlisting,swap_flat` (after) | 23 |

## Tests

- Two new cases covering the mapping and the passthrough of other values
- Updated the full-URL conversion test that previously asserted the broken passthrough
- `TEST_MODE=offline npx vitest run test/services/immoscout/immoscout-web-translator.test.js` → 13/13 green, lint/prettier clean

### DIFF
--- a/lib/services/immoscout/immoscout-web-translator.js
+++ b/lib/services/immoscout/immoscout-web-translator.js
@@ -103,6 +103,13 @@ const EQUIPMENT_MAP = {
   lodgerflat: 'lodgerflat',
 };
 
+// The web UI uses "swapflat", but the mobile API only understands "swap_flat".
+// An unknown value is not ignored: the API silently returns 0 results for the
+// whole search. Other values (e.g. "projectlisting") are identical on both APIs.
+const EXCLUSION_CRITERIA_MAP = {
+  swapflat: 'swap_flat',
+};
+
 const REAL_ESTATE_TYPE = {
   'haus-mieten': 'houserent',
   'wohnung-mieten': 'apartmentrent',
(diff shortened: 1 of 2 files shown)


orangecoding:
Thanks!

PR #335, opened 2026-06-10
fix: use hash-router URL format for listing links in email adapters

### LINKED ISSUE #333: "Open in Fredy" links from email adapters land on the overview instead of the listing
## What happens

Clicking "Open in Fredy" in a notification email opens Fredy, but after login you end up on the overview — never on the listing the email was about.

## Why

The UI is served through a `HashRouter`, so routes only exist after the `#`. Most adapters already build the link correctly:

```js
// telegram, slack, discord, ntfy, mattermost, pushover, apprise, console
`${baseUrl}/#/listings/listing/${id}`
```

But the email adapters (**resend, smtp, mailJet, sendGrid**) and the **http** adapter are missing the `/#`:

```js
// resend.js, smtp.js, mailJet.js, sendGrid.js, http.js
fredyUrl: `${baseUrl}/listings/listing/${l.id}`
```

The router never sees the route in the path part, so it silently falls back to the default view.

## Verified

Headless Chrome against a live instance:

- `https://<host>/listings/listing/<id>` → login → ends on `#/dashboard` (console: `No routes matched location "/login"`)
- `https://<host>/#/listings/listing/<id>` → login → lands on the exact listing, detail page renders

Fix is a one-character-ish change in five adapters, PR incoming.

michiosw:
Fixes #333

## Problem

The UI is a `HashRouter` app, and most adapters already link to `${baseUrl}/#/listings/listing/:id`. The email adapters (resend, smtp, mailJet, sendGrid) and the http adapter build the link without the `/#`, so the router ignores the route and the user lands on the overview instead of the listing.

## Fix

Added the `/#` to the five remaining adapters — they now produce the exact same URL shape as telegram, slack, discord, ntfy, mattermost, pushover, apprise and console.

## Verified

Headless Chrome against a live instance: the hash-format URL goes login → exact listing detail (the post-login redirect keeps the deep link intact), while the old path-format URL ends on `#/dashboard`.

### DIFF
--- a/lib/notification/adapter/http.js
+++ b/lib/notification/adapter/http.js
@@ -14,7 +14,7 @@ const mapListing = (listing, baseUrl) => ({
   size: listing.size,
   title: listing.title,
   url: listing.link,
-  fredyUrl: baseUrl && listing.id ? `${baseUrl}/listings/listing/${listing.id}` : null,
+  fredyUrl: baseUrl && listing.id ? `${baseUrl}/#/listings/listing/${listing.id}` : null,
 });
 
 export const send = async ({ serviceName, newListings, notificationConfig, jobKey, baseUrl }) => {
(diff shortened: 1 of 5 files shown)


orangecoding:
Thanks!

PR #337, opened 2026-06-10
fix(sqlite): prevent path traversal in notification adapter dbPath

andesyteoss:
## Vulnerability summary

The SQLite notification adapter (`lib/notification/adapter/sqlite.js`) accepts a user-supplied `dbPath` field that is used directly in a `new Database(dbPath)` call without any validation. Because `better-sqlite3` creates the database file (and any intermediate directories via `fs.mkdirSync`) at whatever path it receives, a crafted `dbPath` like `../../../../tmp/evil.db` or an absolute path like `/etc/evil.db` lets an attacker write files to arbitrary locations on the filesystem.

**CWE:** CWE-22 (Improper Limitation of a Pathname to a Restricted Directory)  
**Severity:** Medium  
**Affected file:** `lib/notification/adapter/sqlite.js`, function `send()`  
**Data flow:** `request.body.fields.dbPath` → `notificationConfig[].fields.dbPath` → `new Database(dbPath)` + `fs.mkdirSync(path.dirname(dbPath))`

## Exploitation

The `POST /api/jobs/notificationAdapter/try` endpoint is protected by `authHook` but **not** `adminHook` (see `lib/api/api.js:72-73`). Any authenticated user — not just admins — can trigger the SQLite adapter's `send()` function with an arbitrary `dbPath`.

Fredy is a multi-user application. Non-admin users do not otherwise have the ability to write files to the server filesystem, so this vulnerability genuinely expands their attack surface. On bare-metal deployments, this allows arbitrary file writes as the Node.js process user. Inside Docker (the recommended deployment), impact is limited to the container filesystem but still permits overwriting application code for potential code execution.

### Proof of Concept

An authenticated user sends:

```bash
curl -X POST http://localhost:9998/api/jobs/notificationAdapter/try \
  -H "Content-Type: application/json" \
  -H "Cookie: fredy-admin-session=<valid-session-cookie>" \
  -d '{
    "id": "sqlite",
    "fields": {
      "dbPath": {
        "value": "../../../../tmp/traversal-test.db"
      }
    }
  }'
```

On the unpatched code, this creates `/tmp/traversal-test.db` (or the equivalent escaped path) and writes a SQLite database there. The `dbPath` value `../../../../tmp/traversal-test.db` resolves outside the application directory because `path.resolve(process.cwd(), '../../../../tmp/traversal-test.db')` walks up the directory tree.

You can verify the file was created:

```bash
ls -la /tmp/traversal-test.db
```

After applying this patch, the same request returns a 500 error with `"Database path must be within the application directory"` and no file is created.

## Fix description

This PR adds a `safeDatabasePath()` validation function that enforces three rules before the path reaches `better-sqlite3`:

1. **Reject absolute paths** — `path.isAbsolute(raw)` catches `/etc/...` style paths immediately.
2. **Confine to `process.cwd()`** — the path is resolved to an absolute path via `path.resolve()`, then verified to start with the application's working directory. This catches `../` traversal sequences regardless of how they're nested.
3. **Require `.db` extension** — ensures the created file has a database extension, reducing the risk of writing files that other system tools might auto-execute (e.g., cron scripts).

The validation runs at the top of `send()`, before `fs.mkdirSync` or `new Database()` are called, so no filesystem side effects occur on invalid input.

## Tests

Added `test/notification/sqlite.test.js` with coverage for:

- ✅ Simple relative paths within cwd (allowed)
- ✅ Nested relative paths within cwd (allowed)
- ✅ Default path `db/listings.db` (allowed)
- ✅ Absolute paths like `/etc/passwd`, `/tmp/evil.db` (rejected)
- ✅ Traversal sequences like `../../etc/cron.d/evil.db` (rejected)
- ✅ Nested traversal like `db/../../../../../../etc/cron.d/evil.db` (rejected)
- ✅ Non-`.db` extensions (rejected)
- ✅ Integration test: `send()` rejects malicious `dbPath` at runtime
- ✅ Integration test: `send()` rejects absolute `dbPath` at runtime

## Adversarial review

Before submitting, we attempted to disprove this finding. We verified that the notification adapter routes are mounted under `authHook` only (line 71 of `api.js`), not `adminHook` — so the "requires admin" counterargument does not apply. We also checked whether Fastify's body parsing or the adapter lookup logic provides any incidental sanitization of the `dbPath` field — it does not; the raw string flows directly from `request.body.fields.dbPath.value` through `notificationConfig` into `send()`. Docker containerization limits but does not eliminate impact (container filesystem writes can still overwrite application source for code execution). The vulnerability is exploitable by any authenticated user on an unpatched instance.

### DIFF
--- a/lib/notification/adapter/sqlite.js
+++ b/lib/notification/adapter/sqlite.js
@@ -7,6 +7,7 @@ import { markdown2Html } from '../../services/markdown.js';
 import Database from 'better-sqlite3';
 import path from 'path';
 import fs from 'fs';
+import logger from '../../services/logger.js';
 
 /**
  * Resolve and validate the user-supplied dbPath so that the resulting file
@@ -25,6 +26,11 @@ import fs from 'fs';
  */
 function safeDatabasePath(raw) {
   if (path.isAbsolute(raw)) {
+    logger.warn(
+      `[SQLite adapter] Rejected absolute database path: "${raw}". ` +
+        'For security reasons, only relative paths within the application directory are now allowed. ' +
+        'Please update your configuration to use a relative path (e.g., "db/listings.db").',
+    );
     throw new Error(`Absolute database paths are not allowed: ${raw}`);
   }
 
@@ -32,6 +38,11 @@ function safeDatabasePath(raw) {
   const root = process.cwd() + path.sep;
 
   if (!resolved.startsWith(root) && resolved !== process.cwd()) {
+    logger.warn(
+      `[SQLite adapter] Rejected database path that resolves outside the application directory: "${raw}". ` +
+        'For security reasons, only relative paths within the application directory are now allowed. ' +
+        'Please update your configuration to use a relative path (e.g., "db/listings.db").',
+    );
     throw new Error(`Database path must be within the application directory: ${raw}`);
   }
 
@@ -88,7 +99,9 @@ export const config = {
       type: 'text',
       label: 'Database Path',
       description:
-        'Path to the SQLite database file (e.g., db/listings.db). If not specified, defaults to db/listings.db',
+        '⚠️ Security note: The database path must be a relative path within the application directory and must end with .db. ' +
+        'Absolute paths and paths that traverse outside the application directory (e.g., using "..") are rejected. ' +
+        'If you previously used an absolute path, please update it to a relative one (e.g., "db/listings.db").',
       placeholder: 'db/listings.db',
     },
   },
--- a/lib/notification/adapter/sqlite.js
+++ b/lib/notification/adapter/sqlite.js
@@ -8,9 +8,44 @@ import Database from 'better-sqlite3';
 import path from 'path';
 import fs from 'fs';
 
+/**
+ * Resolve and validate the user-supplied dbPath so that the resulting file
+ * always lives under the application's working directory.  This prevents
+ * path-traversal attacks (CWE-22) where a crafted dbPath such as
+ * "../../../../etc/cron.d/evil" could create or overwrite arbitrary files.
+ *
+ * Rules:
+ *  1. Absolute paths are rejected outright.
+ *  2. The resolved path must reside within `process.cwd()`.
+ *  3. The filename must end with `.db`.
+ *
+ * @param {string} raw  The user-supplied database path.
+ * @returns {string}     A safe, resolved absolute path.
+ * @throws {Error}       When the path is invalid or escapes the project root.
+ */
+function safeDatabasePath(raw) {
+  if (path.isAbsolute(raw)) {
+    throw new Error(`Absolute database paths are not allowed: ${raw}`);
+  }
+
+  const resolved = path.resolve(process.cwd(), raw);
+  const root = process.cwd() + path.sep;
+
+  if (!resolved.startsWith(root) && resolved !== process.cwd()) {
+    throw new Error(`Database path must be within the application directory: ${raw}`);
+  }
+
+  if (!resolved.endsWith('.db')) {
+    throw new Error(`Database path must end with .db: ${raw}`);
+  }
+
+  return resolved;
+}
+
 export const send = ({ serviceName, newListings, jobKey, notificationConfig }) => {
   const sqliteConfig = notificationConfig.find((adapter) => adapter.id === config.id);
-  const dbPath = sqliteConfig?.fields?.dbPath || 'db/listings.db';
+  const rawPath = sqliteConfig?.fields?.dbPath || 'db/listings.db';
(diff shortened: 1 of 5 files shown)


lib/notification/adapter/sqlite.js
```diff
@@ -8,9 +8,44 @@ import Database from 'better-sqlite3';
...
+ *  1. Absolute paths are rejected outright.
+ *  2. The resolved path must reside within `process.cwd()`.
+ *  3. The filename must end with `.db`.
+ *
+ * @param {string} raw  The user-supplied database path.
+ * @returns {string}     A safe, resolved absolute path.
+ * @throws {Error}       When the path is invalid or escapes the project root.
+ */
+function safeDatabasePath(raw) {
+  if (path.isAbsolute(raw)) {
```
orangecoding, line 28: With this, you might break already existing installations. Is this really necessary? If so, can you at least please put a warning banner into the sqlite adapter, explaining this situation?

orangecoding (changes requested):

andesyteoss:
Thanks for the review — that's a fair concern.

**Is the validation necessary?** Yes, I believe so. Without it, a user-supplied `dbPath` like `../../../../etc/cron.d/evil` can create or overwrite arbitrary files on the host (CWE-22 path traversal). Since `dbPath` comes from the adapter configuration UI, anyone with access to the Fredy dashboard can exploit this.

**Breaking existing installations:** You're right that this will break installations using absolute paths (e.g., `/data/fredy/listings.db`) or paths that traverse outside the app directory. The default `db/listings.db` continues to work fine, so most users won't be affected.

**Warning banner — done.** I've pushed a follow-up commit that adds:

1. **`logger.warn()` messages** — when `safeDatabasePath` rejects a path, it now logs a clear warning explaining *what* was rejected and *how* to fix it (e.g., "Please update your configuration to use a relative path like `db/listings.db`"), before throwing.
2. **UI description update** — the `dbPath` field description in the adapter config now starts with a ⚠️ security note explaining the new restrictions.
3. **`sqlite.md` breaking-change banner** — the adapter readme now leads with a highlighted blockquote explaining the breaking change and migration steps.

This way, if someone upgrades and their path gets rejected, they'll see the reason and remediation both in the logs and in the UI. Let me know if you'd prefer the warning handled differently.

orangecoding:
@sebastiondev Sorry one more question before I merge this. Have you checked if this works in a docker container? 

In general, I appreciate you fixing this issue, however it comes with a high price because I know a couple of people who intentionally have their db in a totally different spot to not wipe it if they re-deploy. What's your planed solution for this

andesyteoss:
Great questions — both are already covered, but I should have called them out more clearly. Here's the breakdown:

### Docker

Yes — in `cae2abb` I updated both the `Dockerfile` and `docker-compose.yml` to set `FREDY_DB_ALLOW_DIR=/db`, which matches the existing `/db` volume. So the default Docker setup works out of the box with no extra configuration. I also added a symlink (`ln -s /db /fredy/db`) so the default relative path `db/listings.db` resolves correctly inside the container.

### Users with the DB in a different location

That's exactly what `FREDY_DB_ALLOW_DIR` is for. It's a comma-separated environment variable listing directories that are explicitly trusted as valid database locations. For example:

```bash
# bare metal — allow two external directories
export FREDY_DB_ALLOW_DIR=/data/fredy,/mnt/backup/fredy
```

```yaml
# docker-compose — allow /db and a second mount
environment:
  - FREDY_DB_ALLOW_DIR=/db,/mnt/persistent
volumes:
  - ./db:/db
  - /mnt/persistent:/mnt/persistent
```

So the migration path for existing users is: set `FREDY_DB_ALLOW_DIR` to the directory their DB already lives in and they're done — no need to move any files.

This is documented in the `sqlite.md` breaking-change banner and in the UI field description, but I'm happy to add a top-level note in the main README or CHANGELOG if you think that would help with discoverability.

gorgobacka:
I'm wondering, if this is the right way to fix this issue. Even within the application directory allowing any user writing files anywhere sounds problematic. (Could one user override other users databases?)
Why not allowing only the admin user to setup the database path? Then it's also fine, if the path is anywhere else and it does not require new environment parameters.
Or do I understand something wrong here?

orangecoding:
Actually.. that sounds about right to me... I think it makes sense to extract this (and potentially other) settings to an admin only section... 

Or even easier, simply reject changing the db if one is not an admin. 

As for the SQLite adapter, the usage is at 1% (only a couple of thousand of installs) it might be ok to just put a warning message on there

orangecoding:
I'm closing this, as it is to me too dangerous to break existing installs. Also I've made this more secure by making sure only admins can do thi

PR #361, opened 2026-07-06
feat(listings): implement live reload for new listings using SSE

RaminGe:
This PR implements real-time silent live-reloading for the listings overview page. 

Currently when new listings are fetched and stored by background/scheduled jobs, users on the Listings page do not see them unless they manually reload the page. This breaks the seamless real-time feel of the application.

**Solution:**
We leverage Fredys existing SSE architecture to silently reload the listings state in the background.

**Backend**:
  - Automatically dispatches a `listings:new` event through SSE to the user when new listings are added to the database

**Frontend**:
  - Subscribed to the `/api/jobs/events` SSE channel
  - Refetches the listings page silently when a `listings:new` event is received

### DIFF
--- a/lib/FredyPipelineExecutioner.js
+++ b/lib/FredyPipelineExecutioner.js
@@ -98,6 +98,7 @@ class FredyPipelineExecutioner {
       .then(this._filterBySimilarListings.bind(this))
       .then(this._filterBySpecs.bind(this))
       .then(this._filterByArea.bind(this))
+      .then(this._sendNewListingsToUser.bind(this))
       .then(this._notify.bind(this))
       .catch(this._handleError.bind(this));
   }
@@ -367,7 +368,16 @@ class FredyPipelineExecutioner {
   _save(newListings) {
     logger.debug(`Storing ${newListings.length} new listings (Provider: '${this._providerId}')`);
     storeListings(this._jobKey, this._providerId, newListings);
+    return newListings;
+  }
 
+  /**
+   * Broadcast real-time live reload event to user via SSE broker.
+   *
+   * @param {ParsedListing[]} newListings New listings to broadcast.
+   * @returns {ParsedListing[]} The same listings, unchanged.
+   */
+  _sendNewListingsToUser(newListings) {
     if (newListings.length > 0) {
       try {
         const job = getJob(this._jobKey);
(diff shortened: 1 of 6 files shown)


lib/FredyPipelineExecutioner.js
```diff
@@ -366,6 +367,19 @@ class FredyPipelineExecutioner {
   _save(newListings) {
     logger.debug(`Storing ${newListings.length} new listings (Provider: '${this._providerId}')`);
     storeListings(this._jobKey, this._providerId, newListings);
+
+    if (newListings.length > 0) {
```
orangecoding, line 381: Please extract this in it's own step of the pipeline. something like _sendNewListingsToUser

ui/src/components/listings/ListingsOverview.jsx
```diff
@@ -104,6 +104,41 @@ const ListingsOverview = ({ mode = 'all' }) => {
...
+
+  // SSE connection for live listings updates
+  useEffect(() => {
+    const src = new EventSource('/api/jobs/events');
+
+    const onNewListings = (e) => {
+      try {
+        const data = JSON.parse(e.data || '{}');
+        if (data && data.count) {
+          loadDataRef.current();
```
orangecoding: If I understand correctly, you call the loadData function, so you got the new listings via sse, still load everything? What if a user is currently checking something on page 2? This would route the user back to page 1 and could be pretty confusing? How about instead showing a small popover at the top of the current view: "There's new data available, click here to reload"

orangecoding (changes requested):

ui/src/components/listings/ListingsOverview.jsx
RaminGe: Yea I thought about that issue too. It's a bit different tho, if a user is on page 2 (or has filters applied or whatever) it would reload page 2. That still would mean that it changes the page slightly.

I thought a live reload is nicer because it happens automatic, but I get your point.
What if we live reload only if a user is on page 1, otherwise show a toast/popover? Or do you wanna stick to the popover for all cases?

orangecoding: I think having it consistent would be better, so showing a popover..

Thanks

RaminGe: Done, I've added the Banner which is also used elsewhere above the table

orangecoding (approved):
Thanks

PR #367, opened 2026-07-17
feat: add InBerlinWohnen provider

madsbergstroem:
## Summary

- add `inberlinwohnen.de` as a native Fredy provider
- crawl every result page through Fredy's shared CloakBrowser instead of direct HTML fetches
- parse structured Livewire snapshots and reject incomplete or bot-detection responses instead of treating them as empty results
- enrich new listings from supported partner detail pages with safe fallbacks
- scope similarity deduplication by job so one job cannot hide matching listings from another

## Implementation

The provider reuses one shared browser for pagination and optional detail enrichment, preserving Fredy's proxy, timeout, and bot-detection behavior. It normalizes stable partner IDs, canonical allowlisted deeplinks, total rent, size, rooms, address, image, provider details, WBS, and availability.

`fetchDetails` uses explicit selectors for Berlinovo, degewo, GESOBAU, Gewobag, HOWOGE, and WBM. STADT UND LAND and any failed or unsupported detail extraction retain the structured InBerlinWohnen snapshot description.

Similarity hashes now include `jobId`, preserving cross-provider deduplication within a job without suppressing the same listing in another job. Cache hydration, address normalization, hard-delete eviction, and hidden-duplicate handling use the same identity.

## Verification

- `npm run test:offline`: 41 files, 252 tests passed, 1 live-only test skipped
- `npm test -- test/provider/inberlinwohnen.test.js --testTimeout=180000`: 5 live tests passed, 13 offline-only tests skipped
- live crawl: all 27 current result pages plus one current detail page for each of the six enrichable partner domains
- `npm run lint`: passed
- `npm run format:check`: passed
- `npm run build:frontend`: passed

### DIFF
--- a/lib/provider/inberlinwohnen.js
+++ b/lib/provider/inberlinwohnen.js
@@ -189,7 +189,7 @@ function normalize(o) {
     address: normalizeAddress(item.address),
     image:
       typeof item.imagePath === 'string' && item.imagePath.trim()
-        ? new URL(`/storage/${item.imagePath.replace(/^\/+/, '')}`, BASE_URL).href
+        ? new URL(`/img/${item.imagePath.replace(/^\/+/, '')}?q=90&fit=crop&fm=png&dpr=1`, BASE_URL).href
         : null,
     description: description || undefined,
   };
--- a/lib/provider/inberlinwohnen.js
+++ b/lib/provider/inberlinwohnen.js
@@ -5,10 +5,12 @@
 
 import { buildHash, isOneOf } from '../utils.js';
 import { extractNumber } from '../utils/extract-number.js';
+import { load } from 'cheerio';
 /** @import { ParsedListing } from '../types/listing.js' */
 /** @import { ProviderConfig } from '../types/providerConfig.js' */
 
 const BASE_URL = 'https://inberlinwohnen.de';
+const MAX_PAGES = 100;
 
 let appliedBlackList = [];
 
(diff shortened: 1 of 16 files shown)


lib/provider/inberlinwohnen.js
```diff
@@ -0,0 +1,234 @@
...
+/**
+ * Fetch every server-rendered result page so new jobs are not limited to page one.
+ * @param {string} url
+ * @returns {Promise<{id:string}[]>}
+ */
+async function getListings(url) {
+  async function fetchPage(page) {
+    const pageUrl = new URL(url);
+    if (page > 1) pageUrl.searchParams.set('page', String(page));
+    const response = await fetch(pageUrl, { headers: { Accept: 'text/html' } });
```
orangecoding: why do you use fetch instead of puppeteer? Fetch is known to be caught by bot detection pretty easily

lib/provider/inberlinwohnen.js
```diff
@@ -0,0 +1,234 @@
...
+  url: null,
+  crawlContainer: '[wire\\:snapshot*=\'"item":\']',
+  sortByDateParam: null,
+  waitForSelector: 'body',
+  crawlFields: {
+    id: '@wire:snapshot',
+  },
+  normalize,
+  filter: applyBlacklist,
+  activeTester: isListingActive,
```
orangecoding, line 338: would you mind also implementing `fetchDetails`?

lib/services/similarity-check/similarityCache.js
```diff
@@ -55,7 +55,7 @@ export const initSimilarityCache = () => {
   const allEntries = getAllEntriesFromListings();
   const newCache = new Set();
   for (const entry of allEntries) {
-    newCache.add(toHash(entry?.title, entry?.price, entry?.address));
+    newCache.add(toHash(entry?.job_id, entry?.title, entry?.price, entry?.address));
```
orangecoding, line 58: What do you want to accomplish here?

orangecoding (changes requested):

lib/provider/inberlinwohnen.js
madsbergstroem: Good point. I replaced the direct listing-page fetches with the shared CloakBrowser. The provider now declares `requiresBrowser`, receives the job browser in `getListings(url, browser)`, and reuses it across all result pages and detail requests, including the configured proxy and bot-detection handling. Structurally incomplete pages now fail the crawl instead of looking like an empty result.

lib/services/similarity-check/similarityCache.js
madsbergstroem: The goal is to keep cross-provider similarity deduplication within one job while preventing one job from hiding the same listing from another job. Persistent listing uniqueness is already scoped by `(job_id, hash)`, so the in-memory cache now follows the same boundary. I expanded the tests and fixed the related hard-delete path: hydration and eviction include `job_id`, addresses use the same normalization as storage, and deleting an already hidden duplicate no longer evicts the active listing cache entry.

lib/provider/inberlinwohnen.js
madsbergstroem: Implemented. `fetchDetails` reuses the shared browser and has explicit parsers for Berlinovo, degewo, GESOBAU, Gewobag, HOWOGE, and WBM. STADT UND LAND currently exposes no useful narrative detail block, so it keeps the structured snapshot description. Unknown hosts, bot detection, and parser failures always return the original listing. The live smoke test verifies one current page for every enrichable partner domain.

madsbergstroem:
Addressed all three review threads in `8bc8c9c`: shared CloakBrowser pagination, partner-specific `fetchDetails`, and fully tested job-scoped similarity caching. The new upstream workflow runs are currently waiting for maintainer approval (`action_required`), while the equivalent local offline, lint, format, build, and live-provider checks pass. Re-review would be appreciated.

lib/services/jobs/jobExecutionService.js
```diff
@@ -186,7 +186,7 @@ export function initJobExecutionService({ providers, settings, intervalMs }) {
             browser = null;
           }
 
-          if (!browser && matchedProvider.config.getListings == null) {
+          if (!browser && (matchedProvider.config.getListings == null || matchedProvider.config.requiresBrowser)) {
```
orangecoding: Why is this needed? I know you're using claude.. While this is fine, you should tell it not to push in vars/methods for the sake of doing so.

please remove the requiredBrowser again from where you/claude added it. The shared browser instance should be available in your provider whether or not you make use out of it ;) 

The implementation of your provider is already much more complicated than the rest. 
Thanks.

orangecoding (changes requested):

madsbergstroem:
Thanks for the patience on this, and for the clear steer.

And yes, built with Claude, no secret there. Done: I removed requiresBrowser again. The shared browser now just gets passed into getListings, whether the provider uses it or not. Also stripped it from the tests and the type doc.

On the provider being heavier than the rest, fair point. What's left is mostly the pagination and the fetchDetails you asked for. If any part still feels like too much, tell me which and I'll trim it.

orangecoding:
@madsbergstroem The test is failing. When done it should be ok to merge 
image

madsbergstroem:
Fixed in e66729a. The failing test wasn't about the provider logic, it was an import-order issue in the test file: utils.js sets up the offline mock that stubs the browser launch, but it was imported after puppeteerExtractor, so the offline run launched a real CloakBrowser and failed when the binary wasn't cached. Reordered the imports to match the other provider tests like immowelt. I reproduced the CI condition locally without a cached binary and the full offline suite passes (252 tests). The Test workflow shows action_required, so it needs your approval to actually run.

orangecoding:
@madsbergstroem Thanks. will merge

PR #371, opened 2026-07-21
fix(sqlite adapter): prevent path traversal via user-controlled dbPath

andesyteoss:
## Summary

The SQLite notification adapter (`lib/notification/adapter/sqlite.js`) uses the user-controlled `dbPath` field verbatim to create directories and open a database file. Because notification adapter fields are stored per-job by any authenticated user via `POST /api/jobs`, a non-admin user can supply an arbitrary absolute path or a `../`-containing relative path and cause Fredy to:

- create arbitrary directories (via `fs.mkdirSync(dbDir, { recursive: true })`), and
- create/open a SQLite database file at any location writable by the Fredy process (via `new Database(dbPath)`).

This is CWE-22 (Path Traversal / arbitrary file & directory creation). Severity is moderate: it requires an authenticated account, but Fredy explicitly supports multiple non-admin users, and this capability goes well beyond what the UI is meant to expose (writing outside the app directory, potentially clobbering unrelated files, or seeding files into sensitive locations like `~/.ssh/`, cron directories, or a reverse-proxy webroot depending on how the container/host is provisioned).

## Data flow

1. `POST /api/jobs` accepts a job payload including `notificationAdapter[].fields.dbPath` from any authenticated session (see `authHook` in `lib/api`).
2. The job is persisted to the jobs store.
3. On job execution, `send()` in `lib/notification/adapter/sqlite.js` reads `sqliteConfig?.fields?.dbPath` and passes it directly to `path.dirname`, `fs.mkdirSync`, and `new Database()`.

No sanitisation, base-directory containment, or absolute-path rejection existed on this path prior to the fix.

## Fix

Introduce `resolveSafeDbPath()` which:

- treats the configured value as a path relative to `process.cwd()` (the Fredy application directory), defaulting to `db/listings.db` when unset/blank,
- rejects absolute paths outright,
- resolves the final path with `path.resolve` and verifies it is either equal to the base directory or starts with `baseDir + path.sep` (the trailing-separator check is important — a naive `startsWith(baseDir)` would allow `/app-evil` to bypass a `/app` prefix check),
- throws a clear error on violation so the notifier fails loudly instead of silently writing outside the sandbox.

The adapter's field description is also updated so the UI communicates the new constraint.

The change is contained to a single file (28 insertions, 2 deletions) and preserves backward compatibility for the documented, intended usage (`db/listings.db` or other relative paths under the app directory).

## Proof of concept

As any authenticated non-admin user:

```bash
# 1) Log in and grab the session cookie (replace creds).
curl -c cookies.txt -X POST http://localhost:9998/api/login \
  -H 'Content-Type: application/json' \
  -d '{"username":"user","password":"password"}'

# 2) Create a job whose sqlite adapter writes outside the app directory.
curl -b cookies.txt -X POST http://localhost:9998/api/jobs \
  -H 'Content-Type: application/json' \
  -d '{
        "name":"pwn",
        "provider":[],
        "notificationAdapter":[
          {"id":"sqlite","fields":{"dbPath":"/tmp/fredy-pwn/evil.db"}}
        ]
      }'

# 3) When the job next runs (or is triggered), before the fix Fredy calls
#    fs.mkdirSync('/tmp/fredy-pwn', {recursive:true}) and creates the DB file.
#    After the fix, resolveSafeDbPath() throws and no filesystem write occurs
#    outside the application directory.
```

Relative traversal (`"dbPath":"../../etc/cron.d/evil.db"`) is likewise blocked because `path.resolve(process.cwd(), '../../etc/cron.d/evil.db')` lands outside `process.cwd()` and fails the containment check.

## Testing

- Verified the vulnerable path by reading `lib/notification/adapter/sqlite.js` before the change: `sqliteConfig?.fields?.dbPath` flowed unchanged into `fs.mkdirSync` and `new Database`.
- Verified the fix locally by exercising `resolveSafeDbPath()` with:
  - unset / empty → resolves to `<cwd>/db/listings.db` (unchanged default behaviour),
  - `"db/listings.db"` → resolves inside cwd (accepted),
  - `"/etc/passwd"` → rejected (absolute path),
  - `"../../etc/cron.d/x"` → rejected (escapes cwd),
  - `"db/../db/listings.db"` → accepted (normalises inside cwd).
- `git diff master..HEAD --stat` shows the change is isolated to one file.

## Adversarial review

Before submitting we tried to talk ourselves out of this. The endpoint is authenticated, so is it really a boundary crossing? Yes: Fredy documents and implements a distinction between admin and non-admin users, and neither role is intended to have arbitrary filesystem write on the host. We also considered whether the notification adapter is only reachable to admins — it isn't; the job/adapter APIs are gated by the generic session `authHook`, not an admin check. Finally, we considered whether `better-sqlite3` itself normalises or restricts the path — it doesn't; it opens whatever path you give it and will happily create the file. The containment check is the right layer for this fix.

### DIFF
--- a/lib/api/routes/jobRouter.js
+++ b/lib/api/routes/jobRouter.js
@@ -33,30 +33,39 @@ function getAdapterFields(adapterList, adapterId) {
 
 /**
  * Ensures a non-admin user cannot introduce or change privileged notification
- * adapter fields (e.g. sqlite `dbPath`). If the incoming request tries to set
- * such a field, its value is replaced with the stored (admin-set) value, or
- * dropped entirely when no stored value exists. Admins are unaffected, so
- * existing configurations continue to work unchanged.
+ * adapter fields (e.g. sqlite `dbPath`). Returns either:
+ *   { ok: true, notificationAdapter }        — safe to persist
+ *   { ok: false, adapterId, field }          — non-admin attempted to set or
+ *                                              change a privileged field
+ *
+ * Admins are unaffected, so existing configurations continue to work unchanged.
+ * Non-admin requests that omit privileged fields (or echo back the admin-set
+ * value unchanged) are also allowed through.
  */
 function sanitiseNotificationAdapter(notificationAdapter, existingJob, requestIsAdmin) {
-  if (requestIsAdmin || !Array.isArray(notificationAdapter)) return notificationAdapter;
+  if (requestIsAdmin || !Array.isArray(notificationAdapter)) {
+    return { ok: true, notificationAdapter };
+  }
 
-  return notificationAdapter.map((adapter) => {
-    if (!adapter || typeof adapter !== 'object') return adapter;
+  for (const adapter of notificationAdapter) {
+    if (!adapter || typeof adapter !== 'object') continue;
     const privilegedFields = PRIVILEGED_ADAPTER_FIELDS[adapter.id];
-    if (!privilegedFields || !privilegedFields.length) return adapter;
+    if (!privilegedFields || !privilegedFields.length) continue;
 
+    const incomingFields = adapter.fields || {};
     const existingFields = existingJob ? getAdapterFields(existingJob.notificationAdapter, adapter.id) : {};
-    const fields = { ...(adapter.fields || {}) };
     for (const key of privilegedFields) {
-      if (Object.prototype.hasOwnProperty.call(existingFields, key)) {
-        fields[key] = existingFields[key];
-      } else {
-        delete fields[key];
-      }
+      if (!Object.prototype.hasOwnProperty.call(incomingFields, key)) continue;
+      const incomingValue = incomingFields[key];
+      const existingValue = existingFields[key];
+      // Allow the no-op case where the client echoes back the admin-set value
+      // (common when editing a shared job through the UI).
+      if (incomingValue === existingValue) continue;
+      return { ok: false, adapterId: adapter.id, field: key };
     }
-    return { ...adapter, fields };
-  });
+  }
+
+  return { ok: true, notificationAdapter };
 }
 
 function doesJobBelongsToUser(job, request) {
@@ -216,7 +225,12 @@ export default async function jobPlugin(fastify) {
         return reply.code(403).send({ error: 'Sorry, but you cannot change the Status of our Demo Job ;)' });
       }
 
-      const safeNotificationAdapter = sanitiseNotificationAdapter(notificationAdapter, jobFromDb, isAdmin(request));
+      const sanitised = sanitiseNotificationAdapter(notificationAdapter, jobFromDb, isAdmin(request));
+      if (!sanitised.ok) {
+        return reply.code(403).send({
+          error: `You are not allowed to set the "${sanitised.field}" field on the "${sanitised.adapterId}" notification adapter. Please ask an administrator to configure this value.`,
+        });
+      }
 
       jobStorage.upsertJob({
         userId: request.session.currentUser,
@@ -225,7 +239,7 @@ export default async function jobPlugin(fastify) {
         name,
         blacklist,
         provider,
-        notificationAdapter: safeNotificationAdapter,
+        notificationAdapter: sanitised.notificationAdapter,
         shareWithUsers,
         spatialFilter,
         specFilter,
--- a/lib/api/routes/jobRouter.js
+++ b/lib/api/routes/jobRouter.js
@@ -14,6 +14,51 @@ import { getSettings } from '../../services/storage/settingsStorage.js';
 
 const DEMO_JOB_NAME = 'Demo-Job';
 
+/**
+ * Notification adapter fields that touch the filesystem or other sensitive
+ * process-level resources. Non-admin users must not be able to introduce or
+ * modify these fields on a job, otherwise any authenticated user could make
+ * the server open/create files at arbitrary locations writable by the process
+ * (e.g. the sqlite adapter's `dbPath`).
+ */
+const PRIVILEGED_ADAPTER_FIELDS = {
+  sqlite: ['dbPath'],
+};
+
+function getAdapterFields(adapterList, adapterId) {
+  if (!Array.isArray(adapterList)) return {};
+  const entry = adapterList.find((a) => a && a.id === adapterId);
+  return (entry && entry.fields) || {};
+}
+
+/**
+ * Ensures a non-admin user cannot introduce or change privileged notification
+ * adapter fields (e.g. sqlite `dbPath`). If the incoming request tries to set
+ * such a field, its value is replaced with the stored (admin-set) value, or
+ * dropped entirely when no stored value exists. Admins are unaffected, so
+ * existing configurations continue to work unchanged.
+ */
+function sanitiseNotificationAdapter(notificationAdapter, existingJob, requestIsAdmin) {
+  if (requestIsAdmin || !Array.isArray(notificationAdapter)) return notificationAdapter;
+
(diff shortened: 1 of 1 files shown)


lib/notification/adapter/sqlite.js
```diff
@@ -8,9 +8,35 @@ import Database from 'better-sqlite3';
...
+ * creation via path traversal (CWE-22) since any authenticated user can
+ * configure notification adapter fields on a job.
+ */
+const resolveSafeDbPath = (userDbPath) => {
+  const baseDir = process.cwd();
+  const relative = (typeof userDbPath === 'string' && userDbPath.trim().length > 0)
+    ? userDbPath.trim()
+    : DEFAULT_DB_RELATIVE;
+
+  if (path.isAbsolute(relative)) {
```
orangecoding: We had one person in the past who created exactly the same pr and we rejected it, because this change (line 25) can and will break existing installations using this adapter if they have an absolute path.

What's your take on this

orangecoding (changes requested):

andesyteoss:
Thanks for the context — that's a fair objection, and you're right that a hard rejection of absolute paths is a breaking change for anyone who legitimately configured e.g. `/var/lib/fredy/listings.db`. I don't want to force that migration on existing installs.

The reason I still think there's something worth fixing here is the trust boundary rather than the path shape itself: `notificationAdapter[].fields.dbPath` is settable by any authenticated user on `POST /api/jobs` (the route is gated by the generic session `authHook`, not an admin check), so a non-admin can currently make Fredy `fs.mkdirSync` and open a SQLite file at any location writable by the process — `/tmp/x`, an SSH config directory, a reverse-proxy webroot, etc. That's what I was trying to contain.

Given your constraint, some options that preserve existing installs:

1. **Admin-only for absolute paths.** Allow absolute paths only when the job is being created/edited by an admin; non-admins get relative-only. Backward compatible for anyone who set it up as admin (the normal case), and closes the non-admin escalation.
2. **Opt-in allowlist via env var.** e.g. `FREDY_SQLITE_ALLOWED_DIRS=/var/lib/fredy:/data`. Default = cwd only. Existing installs add one env var on upgrade; nothing silently breaks because the error tells them exactly what to set.
3. **Restrict at the API layer, not the adapter.** Leave the adapter permissive (no behavioural change for existing DBs), but reject `dbPath` changes from non-admin users in the jobs route. Zero impact on existing configs; the fix lives entirely in the authorization layer.

My preference would be (3) — smallest behavioural change, doesn't touch the adapter at all, and matches the actual problem (authorization, not path validation). Happy to rework the PR that way, or (1) if you'd rather keep the check next to the adapter. Or if you'd prefer to leave this as-is and treat "any authenticated user can configure adapters" as intended, I'm fine closing the PR — your call.

orangecoding:
> Thanks for the context — that's a fair objection, and you're right that a hard rejection of absolute paths is a breaking change for anyone who legitimately configured e.g. `/var/lib/fredy/listings.db`. I don't want to force that migration on existing installs.
> 
> The reason I still think there's something worth fixing here is the trust boundary rather than the path shape itself: `notificationAdapter[].fields.dbPath` is settable by any authenticated user on `POST /api/jobs` (the route is gated by the generic session `authHook`, not an admin check), so a non-admin can currently make Fredy `fs.mkdirSync` and open a SQLite file at any location writable by the process — `/tmp/x`, an SSH config directory, a reverse-proxy webroot, etc. That's what I was trying to contain.
> 
> Given your constraint, some options that preserve existing installs:
> 
> 1. **Admin-only for absolute paths.** Allow absolute paths only when the job is being created/edited by an admin; non-admins get relative-only. Backward compatible for anyone who set it up as admin (the normal case), and closes the non-admin escalation.
> 2. **Opt-in allowlist via env var.** e.g. `FREDY_SQLITE_ALLOWED_DIRS=/var/lib/fredy:/data`. Default = cwd only. Existing installs add one env var on upgrade; nothing silently breaks because the error tells them exactly what to set.
> 3. **Restrict at the API layer, not the adapter.** Leave the adapter permissive (no behavioural change for existing DBs), but reject `dbPath` changes from non-admin users in the jobs route. Zero impact on existing configs; the fix lives entirely in the authorization layer.
> 
> My preference would be (3) — smallest behavioural change, doesn't touch the adapter at all, and matches the actual problem (authorization, not path validation). Happy to rework the PR that way, or (1) if you'd rather keep the check next to the adapter. Or if you'd prefer to leave this as-is and treat "any authenticated user can configure adapters" as intended, I'm fine closing the PR — your call.

ok. do 3 please. also check for linting issues (See result of this run)

andesyteoss:
Thanks — reworked as option (3).

The check now lives in the jobs route (`lib/api/routes/jobRouter.js`) instead of the sqlite adapter. Summary of the change:

- `lib/notification/adapter/sqlite.js` is reverted to its original behaviour — no path validation, no changes to how `dbPath` is resolved. Existing installs with absolute paths (e.g. `/var/lib/fredy/listings.db`) continue to work unchanged.
- `POST /api/jobs` now runs incoming `notificationAdapter[]` entries through a `sanitiseNotificationAdapter()` helper. For any adapter field listed in `PRIVILEGED_ADAPTER_FIELDS` (currently just `sqlite.dbPath`), a non-admin request has that field replaced with the stored value from the existing job, or dropped entirely if there isn't one. Admins are untouched — they can still set/change `dbPath` freely.

Net effect:
- Admin-created jobs: no behavioural change at all.
- Non-admin creating a new job: `dbPath` is ignored, adapter falls back to the default `db/listings.db`.
- Non-admin editing an admin-owned/shared job: keeps whatever `dbPath` the admin set; can't override it.

The privileged-fields list is a small registry (`{ sqlite: ['dbPath'] }`) so if another adapter ever grows a similar field it's a one-line addition.

Lint/format:
- `eslint .` → clean.
- `prettier --check` → clean (re-ran Prettier on the file).
- `TEST_MODE=offline vitest run` → 247/247 passing.

Pushed as `38c3163` (force-push because I amended the previous commit on this branch to keep history tidy). Happy to split it into two commits if you'd prefer the revert of the sqlite adapter to be a separate commit for review.

orangecoding:
@sebastiondev You latest change lead to a problem. What if a non admin user tries to do that? There's no error message whatsoever so a non admin user will be totally lost as to why his adapter doesn't get saved

andesyteoss:
Good catch — you're right, silently dropping the field is bad UX. Fixed in `c0f0bd5`.

`POST /api/jobs` now returns HTTP 403 with an explicit error when a non-admin request tries to set or change a privileged adapter field:

```json
{ "error": "You are not allowed to set the \"dbPath\" field on the \"sqlite\" notification adapter. Please ask an administrator to configure this value." }
```

Behavioural details:

- **Admins**: no change — can set/change `dbPath` freely.
- **Non-admin, field omitted**: no change — request goes through, adapter uses its default.
- **Non-admin, field echoed back with the same value the admin stored** (common when editing a shared job through the UI, which round-trips the whole payload): allowed, so shared-job edits don't spuriously 403.
- **Non-admin, field set to a new value**: 403 with the message above, so the frontend can surface it and the user knows exactly what to ask an admin for.

Lint/format/tests still clean:
- `eslint .` → clean (lint-staged also ran on push).
- `prettier --check` → clean.
- `TEST_MODE=offline vitest run` → 247/247 passing.

orangecoding (approved):

andesyteoss:
Thanks for reviewing and merging.

orangecoding:
> Thanks for reviewing and merging.

I feel like I have spoken with Claude all the time. The tone and style of answer made me think that 🤓

Issue #359, opened 2026-07-01
[Feature]: More addresses for distance calculation


FlyingT:
### Related Problem

Currently, only one (home-)address can be added to let Fredy calculate the distance

### Proposed Feature

Option to add multiple addresses for home, job of person 1, job of person 2, ... 

### Benefits

See if the distance to both jobs in a relationship are acceptable for the listing 

### Additional Context

love the project, thanks <3
Grüße von jemandem, der quer durch Deutschland ziehen wird D:

orangecoding:
Fixed here:
https://github.com/orangecoding/fredy/pull/372

PR #372, opened 2026-07-23
feat: Allow multiple addresses (#359)

konlanx:
# Description

This PR adds the logic to handle multiple addresses for the distance calculation.

While using Fredy (which is awesome, thanks so much!) I wanted to add multiple addresses, so I could see where an apartment is located respective to different workplaces and such. I saw that there is already an open issue for it (#359), so I decided to implement it.

The old single address is being migrated to a list of one address upon update, so nothing gets lost.

Please let me know if I should make any changes.

## Example screenshots

image
image
image


orangecoding:
@konlanx Thanks for this. Have you considered removing the respective distance from the new jsonb if one decides to remove the address? I have not seen it in your pr.

Edit: I found it. `resetGeocoordinatesAndDistanceForUser`.
But that removes all calculated distances. If a user e.g. has 5 addresses and only removed 1, it is pretty harsh to remove all others as we would have to recalculate the distance for all that is already available. It's not too bad, but also unecessary, WDYT?

konlanx:
@orangecoding Good catch, thanks!

I reworked it so the addresses get passed through and we only recompute what actually needs recomputation.

orangecoding:
@konlanx You removed `runGeoCordTask();` from userSettingsRoute. This most certainly means no recalculation if I get this correctly...? Was that a mistake?

konlanx:
@orangecoding The idea was that `updateDistancesForAddressChange` recomputes all of the already geocoded listings on save and everything else gets done by the geocoding cron.

orangecoding:
@konlanx am I missing something? (Probably)..
If you add a new address and wait for the cron to kick in, it means that a user has to wait potentially for some time to seethe distance..
As this is only a mathematical computation, I think it does not hurt to call it right away?

konlanx:
@orangecoding You're not missing anything. Sorry.

I added the call back in.

orangecoding:
@konlanx ok good. I almost thought I lost my mind ;)

Will review it one last time today and eventually merge it. Thanks!

konlanx:
@orangecoding Sorry for the extra work! I didn't spend much time in the code yet, so this was a bit of a learning curve here 😆 

Thank you for your time 🙂

PR #398, opened 2026-08-03
Add public transit travel time to listing detail page

emnl51:
Add public transit travel time to listing detail page

Shows a badge next to each saved home/work address on a listing's
detail page with the estimated public transit travel time between
that address and the listing, using the Transitous API that is
already used elsewhere in the app for nearby stops and departures.

Calculation is on-demand only, triggered when a listing detail page
is opened - never batched across a list - and cached for an hour,
since Transitous names journey planning as a heavier endpoint than
simple stop/departure lookups.


orangecoding:
Im working on it too ;)

Issue #400, opened 2026-08-05
[Feature]: Allow cycling as transportation method


konlanx:
### Proposed Feature

I would like to see travel time when cycling, as I am primarily using my bike to get around.

<img width="405" height="152" alt="Image" src="https://github.com/user-attachments/assets/e6ff0a12-701e-4478-acf4-2a1cbb70792f" />

### Benefits

People with bikes can see travel times.

konlanx:
Turns out there was simply one option missing from the dropdown in settings. Initially I thought cycling was not yet supported, but it just wasn't selectable. One word fix.

PR #401, opened 2026-08-05
fix(#400): Add missing bike option to dropdown

konlanx:
Added missing `bike` option to dropdown


orangecoding:
Does this also make it selectable on the detail listing?

konlanx:
Commenting on PR's during vacation is illegal and this incident will be reported!

On topic: Yes, this single addition seems to make it work like every other transportation method.

PR #405, opened 2026-08-08
fix(immoscout): prevent base64 corruption of raw map shape polylines

TomTomNavigator:
ImmoScout web search URLs with custom map shapes pass raw Google Encoded Polylines. Base64-decoding them corrupts the polyline bytes into invalid UTF-8 characters (%EF%BF%BD), causing ImmoScout's mobile API to reject requests with HTTP 400 Bad Request: Cannot parse shape / 500 Internal Server Error.

This change passes the raw shape string directly without base64 decoding.


orangecoding:
@TomTomNavigator  the immoscout tests are failing. Please check

PR #422, opened 2026-08-17
fix(immowelt): use document navigation and extract full multi-section description

chbndrhnns:
### Summary of Changes

- **Multi-section description extraction**: Extract all description sections (property description, furnishings/equipment, location, notes) from Immowelt exposé pages and parse embedded `__UFRN_LIFECYCLE_SERVERREQUEST__` state when available.
- **Robust exposé loading**: Use document navigation and fallback handling for exposé enrichment when subresource fetches fail or encounter restrictions.
- **BFF pacing & error handling**: Support adaptive pacing and backoff on rate limiting or 403 responses.
- **Puppeteer process safety**: Add safe guarded access around `browser.process()` in `puppeteerExtractor.js`.
- **Tests**: Added test cases in `test/provider/immoweltNormalize.test.js` and `test/services/immowelt/immoweltBff.test.js`. All 172 test suites pass with `yarn test:offline`.

### Verification
- `yarn test:offline` (172 passed, 2021 tests)
- `yarn lint` (0 errors)

### DIFF
--- a/lib/provider/immowelt.js
+++ b/lib/provider/immowelt.js
@@ -35,16 +35,75 @@ import logger from '../services/logger.js';
 /** @import { ProviderConfig } from '../types/providerConfig.js' */
 
 /**
- * The exposé element holding the full, unshortened description.
+ * The exposé elements holding description parts.
  *
- * The card payload truncates it at 500 characters, which is enough to read but not enough to
+ * The card payload truncates it at 200-500 characters, which is enough to read but not enough to
  * blacklist on: the terms people filter for ("Tauschwohnung", "WBS", "Zwangsversteigerung") are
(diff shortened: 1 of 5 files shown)


orangecoding:
Thanks!

your diagnosis was right, and the multi-section extraction is in master now. I rebuilt it rather than merging directly: the server-state literal is scanned instead of regex-matched (a description containing "); truncated the payload), and the test uses a real captured exposé instead of synthetic HTML. The browser.process() guard came along too. I left out the switch from XHR to page.goto. That reverses a deliberate choice to avoid a full page load plus its own DataDome challenge per listing, and the description fix works without it. Closing as incorporated.

PR #425, opened 2026-08-17
Fix Immowelt encoded search filters

### LINKED ISSUE #424: [Bug]: Immowelt BFF migration rejects encoded locations and ignores construction-year filters
### Bug Description

Upgrading from Fredy 23.3.4 to 26.3.2 breaks an existing Immowelt job created from an Immowelt search URL. The behavior is also present in 26.4.0 because the Immowelt search-model converter is unchanged between those releases.

The saved URL contains a base64url-encoded JSON `locations` value with a place ID and area information, plus `yearOfConstructionMin`. The new BFF translator sends the entire encoded value as a literal place ID, causing the Immowelt BFF to return HTTP 400. Replacing it manually with the raw place ID avoids the 400, but the radius/area information and construction-year filter are then ignored, silently widening the search.

### Steps to Reproduce

1. Start with a Fredy 23.3.4 Immowelt job whose URL has this sanitized shape:
   `https://www.immowelt.de/classified-search?distributionTypes=Buy&estateTypes=Apartment&locations=<base64url-encoded-location-object>&spaceMin=80&spaceMax=105&yearOfConstructionMin=2010`
2. Upgrade the same database to Fredy 26.3.2 or 26.4.0.
3. Run the existing job.
4. Observe HTTP 400 from the Immowelt BFF because the encoded location object is placed in `criteria.location.placeIds`.
5. Replace `locations` with only the raw place ID and run again.
6. Observe that the request can proceed, but the saved radius/area and `yearOfConstructionMin` constraints are not translated.

### Expected Behavior

Existing Immowelt jobs should remain backward compatible and retain all narrowing constraints after an upgrade. The translator should decode the legacy/current encoded location object and preserve its place ID plus radius/geometry, and it should translate `yearOfConstructionMin`.

If a constraint cannot be translated, the job should fail closed instead of running a broader search.

### Actual Behavior

The encoded `locations` value is treated as a literal place ID, so `/serp-bff/search` returns HTTP 400.

Using only the raw place ID avoids that error, but the translator logs that `yearOfConstructionMin` is unsupported and ignores it. The radius/geometry contained in the encoded location object is also lost, so the resulting search is broader than the saved search.

### Screenshots / Logs

Sanitized runtime observations (timestamps, URL values, host details, and identifiers removed):

```text
Immowelt BFF request failed with HTTP 400 after the encoded locations value was used as a place ID.
Immowelt search url uses filter(s) Fredy cannot translate yet: yearOfConstructionMin. Those are ignored, so the search runs wider than the one in your browser.
```

No debug bundle is attached because it contains private runtime and search data.

### Environment

Docker on Linux; upgrade tested from Fredy 23.3.4 to 26.3.2. The same converter behavior was verified in Fredy 26.4.0.

### Additional Context

The regression appears to come from the switch from page scraping to the Immowelt JSON-BFF translator introduced in commit https://github.com/orangecoding/fredy/commit/fe7f4438c8376bc313979fa7abd4a1853d04827e.

Relevant code: https://github.com/orangecoding/fredy/blob/master/lib/services/immowelt/immowelt-search-model.js

The current translator only maps a fixed set of list, number, and enum parameters. It reshapes `locations` directly into `location.placeIds`, without decoding the encoded JSON form, and warns that unknown narrowing filters are ignored. We rolled back to 23.3.4 because accepting the widened result set would change the semantics of the saved search.

frankcai:
## Summary

- decode and validate Immowelt's base64url-encoded location objects
- translate saved radius boundaries to `criteria.location.polylines`
- preserve existing raw comma-separated place IDs and map `yearOfConstructionMin`
- stop malformed, invalid, or unsupported narrowing filters instead of silently widening searches
- continue ignoring only established tracking/view noise and `utm_*` parameters

## Root cause and impact

The JSON-BFF converter treated an encoded location object as a literal place ID, which caused HTTP 400 responses. Replacing it with a raw place ID avoided that error but discarded the saved radius boundary, while `yearOfConstructionMin` was also ignored. This change preserves those narrowing constraints and fails closed when exact translation is not possible.

The converter remains stateless. Job storage, notification behavior, UI schemas, and provider normalization are unchanged.

Fixes #424

## Validation

- focused Immowelt converter tests: 16 passed
- offline test suite: 172 files passed; 2,020 tests passed and 1 skipped
- ESLint and Prettier checks passed
- direct Immowelt provider verification returned HTTP 200 and matched the intercepted request criteria
- isolated `linux/amd64` image built and started healthy with zero restarts

The reproduction and validation evidence is sanitized; no private search URL values, coordinates, identifiers, database content, or runtime credentials are included.

### DIFF
--- a/lib/services/immowelt/immowelt-search-model.js
+++ b/lib/services/immowelt/immowelt-search-model.js
@@ -25,8 +25,6 @@
  * @see lib/services/immowelt/immoweltBff.js for the transport that sends this.
  */
 
-import logger from '../logger.js';
-
 /** Base url of the search page these query strings come from. */
 export const IMMOWELT_SEARCH_PATH = '/classified-search';
 
@@ -43,7 +41,15 @@ export const DEFAULT_ORDER = 'DateDesc';
 const LIST_PARAMS = ['distributionTypes', 'estateTypes', 'estateSubTypes', 'projectTypes', 'featuresIncluded'];
 
 /** Query parameters holding a single number, copied to `criteria` as numbers under the same name. */
-const NUMBER_PARAMS = ['priceMin', 'priceMax', 'numberOfRoomsMin', 'numberOfRoomsMax', 'spaceMin', 'spaceMax'];
+const NUMBER_PARAMS = [
+  'priceMin',
+  'priceMax',
+  'numberOfRoomsMin',
+  'numberOfRoomsMax',
+  'spaceMin',
+  'spaceMax',
+  'yearOfConstructionMin',
+];
 
 /**
  * Query parameters holding a single enum value. The url spells these in whatever case the UI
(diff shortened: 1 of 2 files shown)


orangecoding:
Thanks! this is in master now as a superset. Your radius shape (placeId/polyline) is handled alongside the map-drawing shape ({"drawings":[…]}), and I took your fail-closed treatment of non-numeric ranges. Two deviations: {placeId, radius} without a polyline now throws instead of falling back to the whole place, and encoded entries may still be combined with plain place ids. Also added bbox → location.geometry and yearOfConstructionMax. Closing as incorporated.

Issue #368, opened 2026-07-17
[Feature]: Add native OpenID Connect login


madsbergstroem:
### Related Problem

Fredy currently supports only local username/password login. This requires a separate credential even when an installation already has an OpenID Connect identity provider.

This is related to #362, but proposes native OIDC rather than trusting identity headers from a reverse proxy. Native OIDC keeps authentication inside Fredy's trust boundary and works independently of a specific proxy.

The configurable session TTL also does not currently remove the frequent-login problem completely: the session cookie uses `settings.sessionTTL` in [`lib/api/api.js`](https://github.com/orangecoding/fredy/blob/master/lib/api/api.js#L36), while authorization separately enforces a hard-coded two-hour maximum in [`lib/api/security.js`](https://github.com/orangecoding/fredy/blob/master/lib/api/security.js#L8-L18).

### Proposed Feature

Add optional, standards-based OpenID Connect login using Authorization Code Flow with PKCE.

Suggested first scope:

1. Configure an issuer URL, client ID, client secret, and redirect URL through server-side configuration/environment variables.
2. Use issuer discovery (`.well-known/openid-configuration`) and the provider's authorization, token, JWKS, and optional end-session endpoints.
3. Add a login endpoint that stores cryptographically random `state`, `nonce`, and the PKCE verifier in the server session before redirecting to the identity provider.
4. Add a callback endpoint that exchanges the code and validates the signature, issuer, audience, expiry, state, nonce, and PKCE challenge before creating the normal Fredy session.
5. Identify users by the immutable `(issuer, subject)` pair. Do not use proxy headers or email alone as the authentication key. Account linking or auto-provisioning should be explicit and disabled by default.
6. Keep local password login as a fallback by default, with an option to hide/disable it only after OIDC has been verified.
7. Derive Fredy's hard session expiry from the same `sessionTTL` setting used for the cookie so both checks agree.

Possible endpoints:

- `GET /api/login/oidc`
- `GET /api/login/oidc/callback`
- existing `POST /api/login/logout` destroys the local session and may optionally redirect through the provider's end-session endpoint

### Security and Failure Behavior

- OIDC is opt-in; incomplete configuration must leave the existing password flow unchanged.
- Discovery, token exchange, or validation failures return to the login screen without creating a session.
- Callback state is single-use to prevent replay.
- Client secrets remain server-side and are never returned by the settings API.
- Unknown subjects are rejected unless an administrator explicitly enables a constrained provisioning policy.
[...]

PR #434, opened 2026-08-23
feat(auth): opt-in sign-in from a trusted reverse proxy (forward auth)

### LINKED ISSUE #362: [Feature]: Adding proxy forward auth
### Related Problem

_No response_

### Proposed Feature

I would like Fredy to offer support for Proxy Forward Auth (or alternatively OIDC, which is more complex) . Currently, I have to sign in multiple times a day when opening a listing in Fredy. The goal is to simplify this by offering an alternative auth way.
Alternatively, if the sessions could live longer, this would also be a great enhancement.

### Alternatives Considered

_No response_

### Benefits

Less sign ins.

### Additional Context

If you are open for this feature, I'm happy to create and contribute a PR.

Also, thanks for creating Fredy!

cedricziel:
## Problem

Fredy behind an identity-aware reverse proxy (Pangolin, Authelia, Authentik, oauth2-proxy, Traefik forward-auth) means signing in twice: once at the proxy, once into Fredy. The proxy has already verified who the user is and passes it along in `Remote-User`, but Fredy ignores the header. #362 / #368 asked for this before; I read the reasoning in #368 ("defence in depth rather than a trusted-header contract") and tried to build exactly the opt-in, narrow version that keeps that property for everyone who does not switch it on - and keeps Fredy's own session as the thing that actually authorises requests for those who do.

## Approach

Off by default. When an admin enables it (*Administration → System → Reverse proxy sign-in*):

- A global `preHandler` hook (`lib/api/proxyAuth.js`) looks at requests that have **no valid session**. A valid session always wins, so nobody is silently switched.
- The identity header is accepted **only** from the configured trusted proxy addresses (IPs/CIDRs, IPv4 and IPv6), checked against the **TCP peer** - never `X-Forwarded-For`, which is what an attacker behind the proxy controls.
- Optionally, only when a **shared-secret header** (name + value chosen by the admin, value write-only, constant-time compare) matches as well.
- The header value is mapped onto an **existing** user with that username. Nothing is created, nothing is promoted; unknown names fall through to the normal login form.
- Establishing the session is the same three writes `POST /api/login` does. `sessionTTL`, admin checks, MCP tokens, rate limiting: untouched.
- Settings are validated on save (trusted list must parse and be non-empty when enabled; header names must be RFC 7230 tokens). `proxyAuthSecret` joins `session_secret` in the never-serialised set.
- Trusted-proxy list is parsed once per distinct setting value (memoised), so the per-request cost is a `BlockList.check`.

Header default is `Remote-User` (Pangolin, Authelia); configurable for `X-authentik-username` etc. Secret header name is free-form.

UI: one checkbox plus four dependent fields using the existing `settingsShell__subSettings` pattern, strings in en/de/tr. README gets a section with the setup steps and an explicit warning that every path to Fredy must go through the proxy.

## Tests

- `test/api/proxyAuth.test.js` - CIDR/IP parsing (v4/v6, garbage throws), disabled → null, trusted vs untrusted peer, IPv4-mapped IPv6 peers, empty/unparseable trust list trusts nobody, unknown user → null, configurable header, empty/multi-valued header ignored, shared secret required/mismatch/match, secret header without secret fails closed, memoisation follows a changed setting, settings validation and write-only-secret normalisation.
- `test/api/proxyAuthSession.test.js` - the hook establishes a session, `authHook` accepts it on the first request, feature-off is a no-op, an existing valid session is left alone, an expired session is re-established, unknown identity stays unauthorised, no-session requests are a no-op.
- `test/api/settingsExposure.test.js` - `proxyAuthSecret` never reaches the browser, admin included.
- `yarn test:offline`: 188 files / 2258 tests green. `yarn lint`, `yarn format:check`, `yarn build:frontend` clean.

Unit/integration-tested only so far; I run Fredy behind Pangolin 1.21 and will report back here once the branch image has been through that deployment.

Closes #362


cedricziel:
Verified end to end on a live deployment now.

Setup: this branch built to an image and run as a second, isolated instance behind **Pangolin 1.21.1**. Pangolin authenticates at the edge (SSO), forwards `Remote-User` and adds a per-resource `X-Fredy-Proxy-Secret` header; Fredy configured with the proxy's Docker network as the trusted range and the matching secret.

Results:
- Browser hitting the public hostname → Pangolin login → lands straight on the Fredy dashboard, **no Fredy login form**. Log: `INFO: User <name> signed in via trusted proxy 172.16.x.x`.
- Correct `Remote-User` + secret from the trusted network → `GET /api/login/user` returns the user.
- **Wrong** secret → anonymous (`{}`).
- Same headers from an **untrusted** peer (LAN, outside the trusted CIDR) → anonymous (`{}`).
- Feature left disabled → password login behaves exactly as before.

Happy to adjust naming/shape to whatever you'd prefer if you're open to the feature.

orangecoding:
cool. thanks. keep it coming ;)

PR #435, opened 2026-08-23
fix(mcp): submit response_type from the OAuth consent form

cedricziel:
## Problem

The MCP OAuth authorization-code flow (from #407) can't complete: clicking **Allow** on the consent screen always fails with `{"error":"invalid_request"}`.

The consent form rendered by `GET /api/oauth/authorize` omits a `response_type` field. But `authorizationRequest()` validates `response_type === 'code'` from the request **body** on the `POST` too — and the browser only carries `response_type` in the GET query string, not in the form body it submits. So every real client (Claude, etc.) gets past discovery, registration and the consent page, then dead-ends on the final POST.

Reproduced against a live instance:
- `POST /api/oauth/authorize` with the exact fields the form emits (no `response_type`) → `400 invalid_request`
- same POST with `response_type=code` added → proceeds (`401 login_required` when unauthenticated)

## Fix

`fix(mcp): submit response_type from the OAuth consent form` — add the hidden `response_type=code` input to the consent form so the Allow POST validates. One-line change, plus regression tests asserting the form carries it and that a body without it is still rejected.

## Also: branded consent + sign-in pages

`feat(mcp): brand the OAuth consent and sign-in pages` — the consent screen and the "sign in first" interstitial were unstyled default-serif pages. They now use Fredy's chrome: a centred card with the wordmark, the requesting client's name, the granted scopes as a ticked list, a primary Allow button, and a footer naming the redirect host. Styles are inline/self-contained (renders correctly behind a reverse proxy forwarding only `/api/oauth/*`), follow the OS light/dark preference, and all user-supplied values stay HTML-escaped.

| Light | Dark |
| --- | --- |
|  |  |

## Testing

`TEST_MODE=offline npx vitest run test/api/mcpOAuthRoute.test.js` — 9 passing, including the two new regression tests. Lint/format/copyright hooks pass.


orangecoding:
@cedricziel havent you said you tested it live?

cedricziel:
@orangecoding the forward auth works as expected. - this is a fix to the DCR for MCP that slipped.

cedricziel:
(the fix works as expected when injected into a running docker container ;) )

PR #444, opened 2026-08-28
feat: listing build year and energy class

ChristianGaertner:
I often want to quickly now the state/age of a building, both the year and the energy class are good proxies for that, most portals also show that info, so this PR adds both the storage and the extraction for this (only if pulling the exposé details is enabled).

I hope dropping this PR without creating an issue is fine :)


orangecoding (approved):
Looks good. Thanks

PR #445, opened 2026-08-30
fix(immoscout): decode URL-safe base64 shape parameter

cumbaba:
Immoscout drawn-area (shape) searches were returning 500 from the mobile API, which looked like IP blocking but wasn't. 

Immoscout encodes the shape parameter with a URL-safe base64 variant (+→_), but `toPolyline()` only recognized standard base64, so such values passed through undecoded and the mobile API rejected them. `toPolyline()` now tries Immoscout's alphabet, then base64url, then no substitution, keeping whichever decode yields a valid polyline; verified against the real API and Immoscout's own resolved-query output (`lastSearchApiUrl`).

### DIFF
--- a/lib/services/immoscout/shape.js
+++ b/lib/services/immoscout/shape.js
@@ -8,20 +8,48 @@
  */
 
 /**
- * A `shape` parameter carrying base64 rather than a bare polyline. The web UI writes the padding as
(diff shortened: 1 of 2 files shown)


orangecoding (approved):
Looks good. Thanks!

PR #452, opened 2026-09-10
feat: pace detail requests and refine cloudfront detection

CappyT:
Adds optional pauses between sequential detail requests and fixes a CloudFront false positive.

`detailFetchDelayMs` and `detailFetchJitterMs` default to zero. Each detail request still waits for the previous one to finish; the configured pause limits the request rate when responses are fast. My searches cover a large area, with 400+ listings per provider, so an initial run can require a large batch of detail requests.

Casa.it includes `x-amz-cf-id` in valid search-page state. The header now counts as bot detection only on short documents; HTTP 403/429 and the existing challenge phrases remain unchanged.

The external challenge-solver integration is excluded. #455-458 contain the same removal and retain their provider API/browser paths.

Validation: `yarn lint`, `yarn test:offline`. Tests cover sequential detail pacing and CloudFront false positives.

## AI disclosure

- [x] `ai:assisted` - AI implemented the review follow-up described below.

**Which AI:** OpenAI Codex (GPT-6), with GPT-5.6 Terra agents for code review.

**How much:** The follow-up revisions, added tests, documentation, commit messages and this PR description were produced with AI. Two separate AI agents reviewed the changes; lint, formatting and offline tests were run. This describes the current revision session rather than certifying the provenance of the original submission or claiming a human line-by-line review.

**Why:** To implement the maintainer's feedback, verify provider payload mappings and check regressions across the related PRs.


README.md
```diff
@@ -449,6 +449,16 @@ Residential proxies are a paid service (usually billed per GB, Fredy's traffic i
...
 
+### The other fix: a challenge-solving scrape service
+
+Some portals do not answer a plain request at all, and the provider then has to read a page that sits behind an anti-bot wall (DataDome and the like). Set **`FREDY_CHALLENGE_SOLVER_URL`** to the endpoint of a scrape service that renders such a page for you, and those reads become one POST: Fredy sends `{"cmd": "request.get", "url": ..., "maxTimeout": 90000}` and expects the page back, either as `html` at the top level or under `solution` (`html` or `response`), together with the cookies and the user agent the service earned them with. That is [FlareSolverr](https://github.com/FlareSolverr/FlareSolverr)'s request shape, so FlareSolverr, [TRAWL](https://github.com/germondai/trawl) and anything else speaking it will do. The variable is read at scrape time, so a container restart is all it takes to add or remove one.
```
orangecoding: I don'T really understand the benefit here.. Taking this from the FlareSolverr page:

```
FlareSolverr starts a proxy server, and it waits for user requests in an idle state using few resources. When some request arrives, it uses [Selenium](https://www.selenium.dev/) with the [undetected-chromedriver](https://github.com/ultrafunkamsterdam/undetected-chromedriver) to create a web browser (Chrome). It opens the URL with user parameters and waits until the Cloudflare challenge is solved (or timeout). The HTML code and the cookies are sent back to the user, and those cookies can be used to bypass Cloudflare using other HTTP clients.
````

How would they be able to solve this if Cloak (a way more advanced version of this) is detected?

File: lib/FredyPipelineExecutioner.js [hunk 2]
```diff
@@ -198,8 +200,16 @@ class FredyPipelineExecutioner {
     // is test-shape logic that used to ship in the production pipeline.
     const listingsToEnrich =
       this._maxDetailFetches == null ? newListings : newListings.slice(0, this._maxDetailFetches);
+    const { detailFetchDelayMs = 0, detailFetchJitterMs = 0 } = this._providerConfig;
     const enriched = [];
-    for (const listing of listingsToEnrich) {
+    for (const [index, listing] of listingsToEnrich.entries()) {
```
orangecoding, line 205: Why would it find hundred at once? It's running sequencially.

lib/services/extractor/challengeSolver.js
```diff
@@ -0,0 +1,82 @@
...
+
+import logger from '../logger.js';
+
+/** How long the solver is given to clear a wall. It renders a page and may escalate tiers. */
+export const SOLVER_TIMEOUT_MS = 90_000;
+
+/**
+ * @returns {string|null} the configured solver, or null when there is none
+ */
+export function challengeSolverUrl() {
```
orangecoding: See my question below. I'm not sure how this would help

orangecoding:
Almost all your other pr's b uild on top of this. I however would refrain from using this capture solver. 
1. I do not see the value
2. We might come into a legally problematic area by adding tools to _actively_ bypassing those checks (Cloak is passively doing this)

orangecoding (changes requested):

README.md
CappyT: My experience was with TRAWL rather than FlareSolverr. In my setup, TRAWL returned usable results in cases where the integrated Cloak browser did not. I haven't isolated the cause, so I can't rule out differences in configuration, timing, or simply luck.

That said, I respect your preference for the project's scope. I'm ok to remove the solver integration from this PR and the dependent PRs, keeping the pacing and CloudFront detection fix.

lib/FredyPipelineExecutioner.js
CappyT: I meant hundreds of listings collected in one run, not hundreds of concurrent requests.

My tests cover a rather large geographic area, with multiple providers returning 400+ listings each. On an initial run, that can mean a large batch of new listings needing detail requests.

Currently, each request starts immediately after the previous one and this change adds a configurable pause between them to reduce the request rate. Subsequent runs only fetch details for new listings.

I'll correct the comments to explain this in a better way.

README.md
orangecoding: Thanks, please do so

CappyT: Updated this and the dependent PRs to remove the external solver

orangecoding (approved):
Looks good to me. Thanks

PR #454, opened 2026-09-10
feat(listings): keep the date a portal published an advert

CappyT:
Stores the portal-provided listing date separately from the date Fredy first finds a listing, and reads it from five existing providers:

- Immowelt: `metadata.creationDate`.
- Flatfox: `published`.
- willhaben: `PUBLISHED` (epoch milliseconds).
- Engel & Voelkers: `publishedAt` from the search payload or enabled detail fetch.
- IMAXX: `WebPage.datePublished` from the enabled detail fetch, matched to the listing URL.

These integrations do not substitute update, move-in, or page-cache timestamps for the listing date. Invalid dates and ISO timestamps without an explicit timezone are ignored. Relative dates such as ImmoScout's "vor 3 Tagen" are left unset.

The checked ImmoScout fixtures and a live search response expose only relative publication strings; the checked detail responses have `onlineSince: null`. No absolute publication date was found in those responses.

Migration 42 adds nullable `published_at`. Storage preserves a known date when an incoming value is missing. The listing grid, table, default sorting, detail view and MCP output expose the portal date with `created_at` as the fallback. Existing rows remain unchanged; this PR does not add historical backfill.

This also supplies the storage/UI foundation for #455-458, whose providers set `publishedAt` separately. Some of those portals expose a last-update date; their provider PRs document that distinction.

Validation: `yarn lint`, `yarn test:offline`. Fixture-based tests verify the five field mappings, distinct update dates, timezone handling, invalid/missing values and unrelated JSON-LD entities. Storage tests cover inserts, conflict handling and sorting.

## AI disclosure

- [x] `ai:assisted`

**Which AI:** OpenAI Codex (GPT-6), with GPT-5.6 Terra for code review.

**How much:** AI implemented the follow-up changes, tests and documentation. I approved the proposed diff before it was pushed. Validation included lint and offline tests.

**Why:** To help implement the review feedback and check the related provider changes.

PR #455, opened 2026-09-10
feat(provider): add tecnocasa and tecnorete

CappyT:
Adds Tecnocasa and Tecnorete using their shared server-rendered search and detail payloads. Both providers use ordinary HTTP requests, keep their state per run and deduplicate paginated results.

Includes the shared pacing/CloudFront changes from #452. Detail requests run sequentially with a configurable pause. This PR contains no external challenge-solver integration.

The detail payload supplies description and `last_published_at` as `publishedAt`; #454 adds storage and display for that date.

Validation: `yarn lint`, `yarn test:offline`, including both provider fixtures and the shared network tests.

## AI disclosure

- [x] `ai:assisted`

**Which AI:** OpenAI Codex (GPT-6), with GPT-5.6 Terra for code review.

**How much:** AI implemented the follow-up changes, tests and documentation. I approved the proposed diff before it was pushed. Validation included lint and offline tests.

**Why:** To help implement the review feedback and check the related provider changes.


lib/services/tecnocasa/network.js
```diff
@@ -0,0 +1,383 @@
...
+    // The detail pages are plain requests, so the pipeline is what paces them - at the same gait
+    // the page walk below keeps. Without it a catch-up run of forty new adverts fired forty detail
+    // requests as fast as the network allowed, and the host refused the whole batch.
+    detailFetchDelayMs: PAGE_DELAY_MS,
+    detailFetchJitterMs: PAGE_JITTER_MS,
+    getListings: (url) => getListings(network, url),
+    normalize,
+    fetchDetails: (listing) => fetchDetails(network, listing),
+    activityProbe: checkIfListingIsActive,
+    priceTracking: {
```
orangecoding, line 399: That path goes through puppeteerExtractor (priceTrackingService.js:161), which returns page.content() after Vue mounts. Vue replaces <estate-show-v1 :estate="...">, so the attribute is gone.

Every probe launches a browser, renders a page, records "unknown price". extractPrice has zero test coverage, so nothing catches it. The PR's own downloadFixtures.js comment states the hydration problem, then the config ignores it.

A fix couold be: use probe (plain fetch, like immoscout) instead of extract, or drop priceTracking like idealista did.

lib/utils.js
```diff
@@ -57,15 +57,27 @@ export function getNotificationAdapters() {
...
+ * readme loader - and those shared modules are called while the plugin itself is being evaluated.
+ * Started together, the second plugin can be handed a copy of the shared module that the first one
+ * has not finished evaluating yet, and reading anything off it then throws. A plugin directory
+ * holds a couple of dozen files, so serialising them costs nothing measurable at startup.
+ *
  * @param {string} directory - Absolute path of the plugin directory.
  * @returns {Promise<any[]>} The loaded modules.
  */
-export function loadPluginModules(directory) {
+export async function loadPluginModules(directory) {
```
orangecoding, line 66: The failure this fixes is real, but the explanation isn't. Native Node ESM already guarantees a dependency is fully evaluated before any dependent's body runs, and concurrent dynamic imports of the same module share one evaluation promise, so the "second plugin gets a half-evaluated copy" case can't happen there. I reproduced it: with Promise.all the full offline suite fails 4 out of 5 runs with

```
ReferenceError: Cannot access 'PAGE_DELAY_MS' before initialization
  at createNetworkConfig (lib/services/tecnocasa/network.js:373)
  at lib/provider/tecnorete.js:20
  at async Promise.all (index 20)
```

but the same parallel load under plain Node succeeds 5 out of 5 (node -e "import('./lib/utils.js').then(m => m.getProviders())"). It's a Vite/vitest module-runner dedup issue, not ESM semantics, and it only surfaces now because tecnocasa and tecnorete are the first two providers sharing a module.

Serialising 23 imports at startup costs nothing, so I'm fine keeping the change, but please reword the comment to say it works around the test runner. As written it will send the next person looking for an ESM bug that isn't there. The alternative fix is to not call createNetworkConfig() at module scope in the two provider files.

orangecoding (changes requested):

lib/services/tecnocasa/network.js
CappyT: I switched to a plain HTTP probe, thanks

lib/utils.js
CappyT: Got it, comment reworded. Sorry for the mistake.

orangecoding:
@CappyT Can you please rebase as I merged #458 

Also opn readme, line 205, Subito lost its 🇮🇹 and got pushed out of the country list. Fold it into line 196 as **🇮🇹 Italy** · Subito · Tecnocasa · Tecnorete and delete this line.

CappyT:
@orangecoding sorry for the wait. I've rebased and corrected the README.md file.

PR #456, opened 2026-09-10
feat(provider): add immobiliare.it

CappyT:
Adds Immobiliare.it using the search API, with URL translation for town and polygon searches. Geography lookups resolve town URLs; unsupported searches fall back to the job browser.

Includes the shared changes from #452. The external challenge-solver integration is excluded. If browser rendering cannot return search data, the provider stops that fallback without calling an external service.

The app detail endpoint supplies descriptions and timestamps. Sequential detail requests use the optional pacing configuration. The later of `creationDate` and `lastModified` becomes `publishedAt`, so this is the portal's listing/update date; #454 adds its storage and display.

Validation: `yarn lint`, `yarn test:offline`, including search translation, geography, detail enrichment and provider fixtures. Endpoint details are in `reverse-engineered-immobiliare.md`.

## AI disclosure

- [x] `ai:assisted`

**Which AI:** OpenAI Codex (GPT-6), with GPT-5.6 Terra for code review.

**How much:** AI implemented the follow-up changes, tests and documentation. I approved the proposed diff before it was pushed. Validation included lint and offline tests.

**Why:** To help implement the review feedback and check the related provider changes.


orangecoding:
@CappyT Rebase please.

Also: 
Re-checked today: www.immobiliare.it is still DataDome-403 for me, both the bare homepage and api-next/search-list/listings/, while the android geography API answers 200. casa.it's esapi and idealista's app API both answer me fine, so this is specific to immobiliare's own host. Please confirm a town search works from a normal server before this goes in.

CappyT:
Thanks for checking. I re-tested today and got the same result too: "api-next/search-list/listings/" answers 403 with a DataDome captcha, from a server IP. My own instance hasn't stored a new Immobiliare.it listing since September 10 and logs a 403 on every run, so it has most likely been blocked since then.

The app's search API (`android-imm-v4.ws-app.com/b2c/v1/properties`) is behind DataDome now too. It still answers through a residential IP, but not from a server.

Idk if it's worth to keep this as it "technically works from a residential IP" or if it should be discarded, but that's your call.

orangecoding:
@CappyT  if it is not reliable, I'd remove it (sadly)

The alternative is frustrated user

CappyT:
Well, it has been fun until it lasted then, *sigh*

In all seriousness, I'm closing the PR. Sorry for the bother to go through this.

orangecoding:
@CappyT No need to say sorry. I think you did great things for Fredy and I would love to see more. The rest of your pr's still work no?

CappyT:
@orangecoding for now, i see no problems on the other providers, fingers crossed.

